PR #11, opened 2022-05-18
Realm system, portals, refactoring of the classroom world management system

lukasgolson:
The tutorial system is still in the works, but I believe that the realm system and world management code written to help us, instance players, among tutorials is stable enough to be merged into development. 

The realm system has been created to help us support instanced player tutorials (e.g., private tutorials on the server), improve the resilience of the classroom system, and reduce technical debt. All features have been implemented as generally as possible so that realms can apply to any world modification. Any feature ideas or feedback as to where the API can be improved, simplified, or expanded is greatly appreciated. 

This PR adds a realm system, portals, and refractors world management code to be more resilient and easy to expand.

The realm system divides the map into areas of 8 X 8 X 4 map chunk size (80 nodes per map chunk) with borders 4 map chunks large. It should be noted that realms do not have to take up the entire realm area. Realms are managed by the system and can be easily created or deleted (and associated blocks asynchronously cleared). Realms have helper functions to generate basic floors, invisible border blocks and manage schematics.

Realms are currently created along the X and Z axis, but the system can very easily be expanded to support placement along the Y-axis as well.  Realms are tightly integrated with schematics. Realms can be generated from schematic files, and can be very easily exported to schematic files (currently a single test command that exports the realm and needed meta-info). By using the Lua Voxel Manipulator, realm block manipulation (e.g., creation, deletion, generation) is very fast and uses minimal system resources. Asynchronous code is used from realm deletion to prevent crashing the server.

The system spawns players in a default "spawn" realm. Allowing us to use the single-node generator type and prevent missing block errors. 

The portal framework from the Nether Mod has been integrated and heavily modified for easy player transport and developer ease of use. By calling a single function, developers can easily create realms that are accessible via portals. These realms are created on-demand as players traverse them and can either be shared among all players or instanced (so that each player has a copy of the realm). Portal blocks are dynamically created, hardware coloured based on realm name.

The classroom system has been refactored to use the realm system, improving code resilience and providing a unified API for working with the world. 

**Future additions might include:**
- Automatically registering new schematics exported from realms for new classrooms, portals, etc.  
- Exposing everything to config files (such as world schematic types, portal realms, player instancing).
- Adding in-game menus to manage config files.
- Recycling realm locations after they're deleted by adding realm ID to a list.
- Integration with a custom world generator to include terrain generation from any existing world generator.

All code is heavily documented, but if there are any questions or concerns, please let me know so that I can quickly resolve the issue/concern.


mods/mc_teacher/gui_dash.lua
```diff
@@ -659,414 +657,348 @@ end
...
+            show_tasks(player)
+        elseif fields.lessons then
+            show_lessons(player)
+        elseif fields.classrooms then
+            show_classrooms(player)
+        elseif fields.players then
+            show_players(player)
+        elseif fields.mail then
+            show_mail(player)
+        end
```
Lowie375, line 689: The code for handling the `fields.species` conditional (from `magnify`) seems to have been lost while resolving merge conflicts
```suggestion
        elseif fields.species then
            show_species(player)
        end
```

mods/mc_teacher/gui_dash.lua
```diff
@@ -659,414 +657,348 @@ end
...
+            -- TODO: eventually delete this button, only useful for testing
+        elseif fields.deleteall then
+            pmeta:set_string("classrooms", nil)
+            minetest_classroom.classrooms:set_string("classrooms", nil)
+            show_classrooms(player)
+        else
+            -- escape without input
+            return
         end
     end
```
Lowie375, line 871: The code for handling the plant compendium formspecs (from `magnify`) also seems to have been lost while resolving merge conflicts

```suggestion
    end
    
    if formname == "mc_teacher:species_menu" then
		if fields.back then
		  	show_teacher_menu(player)
		elseif fields.species_list then
        	local event = minetest.explode_textlist_event(fields.species_list)
        	if event.type == "CHG" then
        		context.species_selected = event.index
        	end
		elseif fields.condensed_view or fields.expanded_view then
			if context.species_selected then
      			local ref = get_species_ref(context.species_selected)
          		local full_info = magnify.get_species_from_ref(ref)

				if full_info ~= nil then
          			if fields.condensed_view then -- condensed
						minetest.show_formspec(pname, "mc_teacher:species_condensed", get_condensed_species_formspec(full_info.data))
            		else -- expanded
            			minetest.show_formspec(pname, "mc_teacher:species_expanded", get_expanded_species_formspec(full_info.data, full_info.nodes, ref))
            		end
				else
					minetest.chat_send_player(pname, "An entry for this species exists, but could not be found in the plant database.\nPlease check your server's plant database files to ensure all plants were registered properly.")
				end	
        	end
		end
	end
	
	if formname == "mc_teacher:species_condensed" then
      	if fields.back then 
        	show_species(player)
	    end
    end
    
    if formname == "mc_teacher:species_expanded" then
      	if fields.back then
          	show_species(player)
      	end
    end
```

PR #22, opened 2022-05-27
Removed "pointable" from the barrier node

lukasgolson:
Makes it so that the barrier block behaves more like air -- preventing players from using the barrier block to place nodes and removes the node selection that usually shows up.


pauldpickell:
Nice addition

PR #28, opened 2022-05-29
Refactor tool registration + give/take handing

Lowie375:
PR for merging non-`magnify`-related changes originally developed on the `magnify` branch to `develop`
If this PR is rejected, any refactoring and changes in this PR that were made in `magnify` will be rolled back
* Add code to automatically give/take tools from players when they receive/lose permissions
* Abstract out tool names into local variables


pauldpickell (changes requested):
I think we should change the `check_perm` to `interact` instead of `shout`. I realize this would need to be modified across `mc_student` as well (easy), but if a `teacher` revokes `shout` because a player is spamming the chat, then it would basically make the rest of the game play inoperable. We could also introduce an alternative `student` priv that must be earned/unlocked through tutorials, so that it is unrelated to other basic game play and server management.

Lowie375:
Yep, that makes sense! It seems like `interact` is needed for tools anyway, so we could have `check_perm` check for both `interact` and `student`.
Should we apply a similar change the the magnifying glass (which also uses `shout`)?

pauldpickell:
@Lowie375 Yes, I think all tools should check for `interact`.

PR #31, opened 2022-05-30
Features/tutorials/node destruction

lukasgolson:
The initial "Break-A-Block" tutorial that started the realm system, plus supporting infrastructure to support it. This fully scripted tutorial introduces players to node destruction and tool usage. Formspecs and a HUD are used to guide players through towards completion.

In this branch, the tutorial can be accessed via portal; but it will also be integrated into the tutorialBook on that branch. 

The tutorial scripts start when a player teleports into the tutorial realm; and I'm looking to abstract the tutorial progression system to make it easily re-usable for other types of tutorials. 

To support the tutorial, infrastructural overhauls (some already merged) have been made to the realm system to keep track of which realm players are in, file (and eventually UI) configurable realm callbacks (e.g., when a realm is created or destroyed, when players teleport into / out of realms), and realm meta storage.


pauldpickell:
@lukasdragon Can you check the conflicts in `mc_helpers/init.lua` and `mods/mc_tutorial_framework/init.lua`? These are substantially different from `develop` so not sure if this is desired or if you need to PR `develop` into the `Features/Tutorials/NodeDestruction` branch first.

lukasgolson:
> @lukasdragon Can you check the conflicts in `mc_helpers/init.lua` and `mods/mc_tutorial_framework/init.lua`? These are substantially different from `develop` so not sure if this is desired or if you need to PR `develop` into the `Features/Tutorials/NodeDestruction` branch first.

I'll get that done right now.

PR #32, opened 2022-05-30
Tutorialbook

jocelynbachmann:
Still finishing up the teleport feature as there's an issue with loading schematics, but in a testable state.


mods/mc_tutorial_framework/init.lua
```diff
@@ -2,5 +2,150 @@ mc_tutorialFramework = { path = minetest.get_modpath("mc_tf")}
...
+		table.insert(descriptions, #descriptions + 1, description)
+
+		if schematic then
+			table.insert(schematics, #schematics + 1, schematic)
+		end
+end 
+
+addTutorial("Introduction", "Welcome to Minetest Classroom! To access tutorials, select the topic you would like to learn about on the left. Tutorials can also be accessed via portals that will teleport you to the tutorial relevant to the area you are in. To use a portal, stand in the wormhole until it transports you to a new area. Once you are in the tutorial realm, you can use the portal again to return to the area you were previously in.")
+addTutorial("Test", "testing", "testSchematic")
+addTutorial("Movement", "This tutorial explains how to walk in different directions, jump, and fly. To enter the tutorial, press the 'Teleport to Tutorial' button below. Once you are in the tutorial realm, you can use the portal again to return to the area you were previously in. If you need a reminder on how to use portals, go to 'Introduction'.", "movementTutorial")
```
pauldpickell, line 62: This makes me think that we might need to dynamically issue privs (like 'fly') based on the realm. I can foresee teachers not wanting students to be flying around if the task is to dig or look at the ground. Through the Manage Students formspec, teachers can grant/revoke any priv, but we might want to change default privs on-join.
Edit: @lukasdragon for visibility

mods/mc_tutorial_framework/init.lua
```diff
@@ -2,5 +2,150 @@ mc_tutorialFramework = { path = minetest.get_modpath("mc_tf")}
...
\ No newline at end of file
+mc_realmportals.newPortal("mc_tf","tf_testRealm", false, "testSchematic")
+mc_realmportals.newPortal("mc_tf","tf_movementRealm", false, "movementTutorial")
+
+
+----------------------------------
+--    TUTORIAL BOOK FUNCTIONS   --
+----------------------------------
+
+-- Check for shout priv
```
pauldpickell, line 15: In magnify, we decided to change this priv check to `interact` instead of `shout` because players might be muted for any number of reasons, but we would still want them to be able to do things and complete tasks. Please update this function.

mods/mc_tutorial_framework/init.lua
```diff
@@ -2,5 +2,150 @@ mc_tutorialFramework = { path = minetest.get_modpath("mc_tf")}
...
+			local textarea = mc_tf_menu[5]
+			textarea = "textarea[5,0.2;7.8,8;text;;" .. descriptions[event.index] .. "]"
+			mc_tf_menu[5] = textarea
+			show_tutorial_menu(player)
+		end
+	end
+
+	if fields.teleport then
+		if event.index ~= 1 then
+			selectedRealm = mc_realmportals.CreateRealmByName(names[event.index + 1], schematics[event.index - 1])
```
pauldpickell, line 126: Is this efficient, even if we recycle coordinate space, to be creating a new realm every time a player needs to access a tutorial? Perhaps I missed it, but I did not see anything about managing these realms when the player has left. It would be good to have the opportunity to integrate some tutorials on campus, but this would be complicated by the need to create new realms or even handle it as a realm. Surely for some of these tutorials we could have a static/ dedicated realm or campus location.
Edit: @lukasdragon for visibility

pauldpickell (approved):
@Lowie375 has changed the `check_perm` from `shout` to `interact` in `mc_students` in this [commit](https://github.com/ubc-minetest-classroom/Minetest_Classroom/pull/34/commits/b35a37b0718d0efe144f00e40779bfc92549a39f) of a draft PR #34. Please ignore that comment. 

Should consider my other comments about player instancing for tutorial realms. I think we should support the case where the teacher wants to set the spawn point for the tutorial in an existing (large) realm, like UBC campus.

mods/mc_tutorial_framework/init.lua
lukasgolson: This should be a relatively quick addition. Most of the infrastructure necessary to facilitate this has already been added. I'll get started on that soon.

mods/mc_tutorial_framework/init.lua
lukasgolson: Can be efficient. Creating a new realm with schematics should be faster than generating new chunks using the terrain gen system. 

In terms of coordinate space property, creating a new realm for each tutorial instance is not efficient right now; will be decent soon. Being able to toggle whether it creates a new realm every time a player needs access or not would be good. The portal system currently works that way. Some tutorials (such as the node destruction one) needs to be reset every time it starts, whereas movement does not. 

I've been thinking about the best way to recycle coordinate space, and how we can recycle them efficiently. I'm trying to figure out something that takes minimal resources but is also flexible/resilient in case the server crashes.

For tutorials on campus, the following might work. We could use a table with selectedRealm in the book; On the table could be a function that is used when the player clicks on the "teleport" button. When we add new tutorials, we could supply a flag to define whether we teleport to a realm, coordinates,  open a formspec, etc.,

pauldpickell: > Some tutorials (such as the node destruction one) needs to be reset every time it starts, whereas movement does not.

I am thinking about the `Your Land` server (your-land.de:30000) that has a series of tutorials in the same building. The blocks are reset using a `mesecons_button`, probably triggering the placement of a schematic and then turning the player POV 180 degrees to stop the player from further interaction. This server does not use the `protector` mod, which is interesting. There is also a working example of giving players certain tools by hitting a `mesecons_button`, which could be a nice way to give players a clinometer, compass, etc.

PR #33, opened 2022-05-30
Remove unnecessary call to `magnify`

Lowie375:
Removes an artifact from the old plant compendium implementation that was reintroduced while merging recent changes from `magnify` - this artifact coupled `magnify` and `mc_teacher`


Lowie375:
Superceded by #34

PR #34, opened 2022-05-30
Update magnifying tool privileges + decouple `magnify` and `mc_teacher`

Lowie375:
* Change magnifying tool privilege from `shout` to `interact`
* Handle #33 (removes an artifact that coupled `magnify` and `mc_teacher`)


Lowie375:
Superceded by #35

Issue #40, opened 2022-05-31
Fix crash when teleporting to realm for first time


Lowie375:
Triggered by using `/realmTP 1` in chat for the first time - subsequent uses worked fine

Local error log from `debug.txt`:
```
2022-05-31 12:13:48: ERROR[Main]: ServerError: AsyncErr: Lua: Runtime error from mod 'mc_worldmanager' in callback on_chat_message(): ...ify\mods\mc_worldmanager/realm/realmPlayerManagement.lua:5: attempt to index local 'oldRealm' (a nil value)
2022-05-31 12:13:48: ERROR[Main]: stack traceback:
2022-05-31 12:13:48: ERROR[Main]: 	...ify\mods\mc_worldmanager/realm/realmPlayerManagement.lua:5: in function 'TeleportPlayer'
2022-05-31 12:13:48: ERROR[Main]: 	...test_Classroom-magnify\mods\mc_worldmanager/commands.lua:105: in function 'func'
2022-05-31 12:13:48: ERROR[Main]: 	...netest\minetest-5.5.0-win64\bin\..\builtin\game\chat.lua:79: in function <...netest\minetest-5.5.0-win64\bin\..\builtin\game\chat.lua:52>
2022-05-31 12:13:48: ERROR[Main]: 	...st\minetest-5.5.0-win64\bin\..\builtin\game\register.lua:425: in function <...st\minetest-5.5.0-win64\bin\..\builtin\game\register.lua:409>
```

Seems like this is happening when `oldRealm` hasn't been instantiated (i.e. the first time the command is called in a world). Would recommend using a fallback value or checking for `nil` in the functions that handle teleportation

lukasgolson:
Is this in a new or old world? I'm unable to reproduce this on my end. I'll add in a nil check;

Lowie375:
This was happening in an old world - that may have been the issue
EDIT: Confirmed that this does not happen in newly-generated worlds

pauldpickell:
Fixed by #43

PR #43, opened 2022-05-31
Nil check to stop crashes when teleporting to a new realm for the fir…

lukasgolson:
Hotfix for issue #40


Lowie375 (approved):
Tested the hotfix on a copy of the world that was missing `oldRealmID` and it did not crash! Thanks for the quick patch

PR #48, opened 2022-06-01
Move species viewer formspecs into player inventory

Lowie375:
* Make species viewing formspecs (standard + technical) accessible directly from player inventory, allowing players to return back to the plant compendium more easily
* Refactor `magnify.get_species_from_ref` to return two tables instead of one table with two sub-tables


Lowie375:
The `magnify` branch was updated, but this PR doesn't seem to want to update to reflect that. I'll close this PR and open a new one once `magnify` is ready to be merged into.

Lowie375:
Superceded by #58

Issue #1, opened 2022-05-09
Feature: Tutorial framework


lukasgolson:
TODO: Fill out this issue stub

lukasgolson:
It would be nice to make a extendable tutorial framework, where different tutorials can be selected from a tutorial book menu.

Tutorials should spawn players in an instanced location (similar to classrooms), and teach students about different features in the UBC MineTest Classroom.

Tutorials should initially include movement, MineTest features (punch a block), and chat. Framework should be easily expandable to include tutorials for Minetest Classroom features as well as individual mods.

Tutorials should reduce work load required for TA's and professors for teaching students how to use the platform. Might be expandable to include professor defined lesson modules.

pauldpickell:
I like the idea of using the existing classrooms framework to spawn players into the tutorials. This would also allow us to distribute the game with these pre-built tutorial classrooms that would solve the issue of spawning directly into a singlenode mapgen.

jocelynbachmann:
If using classrooms could overcomplicate things, these are some alternative options based on tutorials I've come across on other servers:

1. Use a building in the spawn location with different floors corresponding to different types of tutorials (e.g. floor 1 for basic controls, floor 2 for specific tools, etc.). Instructions would be on the walls or we could use a mob. This is the easiest option, and could work well given there's already a feature for returning the spawn location, so players could return to the tutorial at any time.

2. Use portals or travel boxes to transport to underground (or above ground) tutorial locations specific to the area the player is in. This would probably take a little more time or research, but there's worlds that use portals like this pretty extensively. There seem to be ways to limit who can access portals, which could be useful to restrict certain tutorials to teachers only or to control the pace at which students are uncovering new features. If not for tutorials, portals could still be an interesting mod to look into for teacher use (e.g. switching between groups of students/activities quickly, if this is time consuming using classrooms).

3. Access tutorials via the notebook menu which transport you to a specific tutorial area (or place you in a classroom). This is  probably the most intuitive option for players, but could be complicated and have limited scalability.

Options 2 and 3 potentially require an initial pop up when joining the server to explain how to do basic things like use the notebook or use a portal.

We've refactored the classroom system to make it easier to create worlds (which could be used to implement options 2 or 3). We're currently exploring each of the options.

lukasgolson:
[...]

PR #61, opened 2022-06-03
Tutorialbook updates

jocelynbachmann:
All previous bugs with teleporting and blank schematics should be fixed. The movement tutorial still has signs that don't retain text, but that's being fixed on Features/Tutorials/MovementTutorial.


pauldpickell:
[See my comment](https://github.com/ubc-minetest-classroom/Minetest_Classroom/issues/1#issuecomment-1146204577) about your refactoring idea.

pauldpickell (changes requested):
Please submit this PR to `develop` instead of `main`.

pauldpickell (changes requested):
The directory `display_modpack-master` should be renamed to `display_modpack` to avoid dependency and loading issues.

PR #69, opened 2022-06-06
Feature/variable realm size

lukasgolson:
Finished variable realm size;

This PR:
1) Marks the unused space when creating new bins as empty
2) Reduces the amount of memory needed to keep track of empty world space by consolidating the list of empty chunks at server startup; 
3) supports bin resizing within gaps left by deleting realms (by-product of point 2).

The current iteration of the code is slow by nature of what it does with nested loops. That said, in testing with the creation (and subsequent deletion) of 10,000 realms, it was still very fast (took about 0.86 seconds on a Ryzen 3700x desktop; 2 seconds on an older mid-tier Intel I7 laptop) and reduced overall memory footprint. 

In the future, and depending on actual usage (such as if we have this code run more often), we might want to introduce more complicated algorithms to try to speed up the process, but for now, this is far more than sufficient.

### DIFF
--- a/mods/mc_worldmanager/commands.lua
+++ b/mods/mc_worldmanager/commands.lua
@@ -162,3 +162,17 @@ minetest.register_chatcommand("realmSetSpawn", {
         return true, "Updated spawnpoint for realm with ID: " .. param
     end,
 })
+
+minetest.register_chatcommand("realmCleanup", {
+    privs = {
+        interact = true,
+    },
+    func = function(name, param)
+        Realm.consolidateEmptySpace()
+        Realm.SaveDataToStorage()
+        return true, "consolidated realms"
+    end,
+})
+
+
+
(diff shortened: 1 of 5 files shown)


mods/mc_worldmanager/commands.lua
```diff
@@ -161,4 +161,18 @@ minetest.register_chatcommand("realmSetSpawn", {
 
         return true, "Updated spawnpoint for realm with ID: " .. param
     end,
-})
\ No newline at end of file
+})
+
+minetest.register_chatcommand("realmCleanup", {
+    privs = {
+        interact = true,
```
pauldpickell: Probably want to consider setting this to `{teacher = true}` instead of `interact`.

lukasgolson: Good catch; I'll update those right now and push again

lukasgolson: Updated!

PR #73, opened 2022-06-09
Features/ubc spawn

lukasgolson:
**A bunch of changes and improvements to allow the realm system to support:**
- The UBC map! (dependent on: https://github.com/pauldpickell/UBC_Campus_Minetest_Map/pull/1)
- Procedural maps as defined by the config file (config files no longer have to correlate with schematics; e.g., how the UBC Minetest map works)
- Better realm command system (commands are now all sub-command based; I still need to write documentation though)
- World-edit schematic format support [to support meta data storage in map schematics]
- Exschem schematic format support (lag free placement; works like the region files in Minecraft)
- Exschem + world-edit (world-edit format managed by exschem) format support
- Support for the legacy (old) schematic format [better than both options above for small realms, but horrible for large realms]
- Bug / crash fixes
- Added a function for realms to "clean" (read: delete) unknown blocks from the realm [highly destructive; necessary for world-edit schematics. Only runs when saving world-edit based schematic files]
- Stopped AFK kicker from kicking while the spawn realm is being placed
- A bunch of new commands to do stuff like manually configure realm location, change the realm used for spawn, etc.,


mods/afkkick/init.lua
```diff
@@ -15,65 +15,72 @@ local players = {}
...
 
-	--Check for inactivity once every CHECK_INTERVAL seconds
-	checkTimer = checkTimer + dtime
 
-	local checkNow = checkTimer >= CHECK_INTERVAL
-	if checkNow then
-		checkTimer = checkTimer - CHECK_INTERVAL
-	end
+    -- Optional dependency on mc_worldManager so we don't kick players while spawn is generating
+    if (mc_worldManager ~= nil and mc_worldManager.SpawnGenerated() ~= nil) then
```
pauldpickell, line 42: Don't you need `teacher` priv to generate spawn? Teachers are not kicked by `afkkick`. Is there another place we need to ensure that the proper player on_join is given `teacher`?

mods/mc_worldmanager/realm/realmSchematicSaveLoad.lua
```diff
@@ -16,15 +17,49 @@ function Realm:Save_Schematic(author)
         fileName = fileName .. math.random(0, 9)
     end
 
-    fileName = fileName .. os.date(" %Y%m%d %H%M")
-
-    local filepath = folderpath .. "\\" .. fileName
-
-    minetest.create_schematic(self.StartPos, self.EndPos, nil, filepath .. ".mts", nil)
+    fileName = fileName .. math.random(0, 99)
```
pauldpickell, line 20: What is the rationale for this file naming convention over the prior date format?

mods/afkkick/init.lua
lukasgolson: Spawn is initially generated before any privileges can/are assigned to a player. This isn't an issue with small spawns (like the current vancouver_osm), but during testing I had the world auto-generate spawn as UBC. Because schematic placement  locked up chat commands (single thread shenanigans), players get autokicked before the player can grant themselves the command. This change was mainly for singleplayer to prevent the spawn from not fully spawning on first launch.

mods/mc_worldmanager/realm/realmSchematicSaveLoad.lua
lukasgolson: We can revert this change; It's no longer necessary -- I forgot to change it back. The rational for the initial change, however, was:
1) Files were being overwritten when saving within the same minute; I could've also changed the filename logic to:
`fileName = fileName .. os.date(" %Y%m%d %H%M%S")`
2) When trying to figure out a rare bug when placing certain schematics, I was generating many randomized realms to schematics. I was finding it difficult to read the time stamps quickly when identifying files.

pauldpickell (commented):
The `ubc_map` submodule here is pointing to your fork, which does not allow me to initialize via git bash and consequently, the folder is empty.

PR #78, opened 2022-06-13
Revert "Features/ubc spawn"

pauldpickell:
Reverts ubc-minetest-classroom/Minetest_Classroom#73 Cannot initialize the `ubc_map` submodule.


lukasgolson:
I'll update the submodule tomorrow morning. Shouldn't take more than a few moments. Sorry for forgetting to update it once the main repo integrated my PR.

Feature should be independent of the UBC map and should work regardless of whether it's installed or not. I'll debug the dependency in the morning.

--update: I've been able to pull the repo and sub-modules anonymously using the Tower git client. I'll try to replicate the issue in Git bash.

PR #79, opened 2022-06-13
Measuring Tape

jocelynbachmann:
Still fixing some bugs with auto-reset and diagonal measurement, but measuring tape is in a testable state and shows a marker of where the tape is laying between start and end points. Aesthetics are temporary – a model is needed to make the tape signlike (so it can be placed on top of the ground/objects instead of temporarily replacing the nodes).


pauldpickell:
@jocelynbachmann can you please send this to `develop` instead of `main`?

Issue #65, opened 2022-06-03
Include UBC as the spawn realm


lukasgolson:
With PR #64, it's now possible to support realms of any size.

I'll configure the UBC schematic(s) to be placed in a spawnRealm.

pauldpickell:
Assets are [here](https://github.com/pauldpickell/UBC_Campus_Minetest_Map).

Edit: Just a note that when spawn realm schematic selection and/or mapgen options are introduced on initiation of the world, we need to update the "Go Home" [button](https://github.com/ubc-minetest-classroom/Minetest_Classroom/blob/df4c0c806e8199333bbca5c3f435f587eb8411b6/mods/mc_student/init.lua#L286-L297):
```
                if fields.spawn then
                        -- TODO: dynamically extract the static spawn point from the minetest.conf file
                        -- local cmeta = Settings(minetest.get_modpath("mc_teacher").."/maps/"..map..".conf")
                        -- local spawn_pos_x = tonumber(cmeta:get("spawn_pos_x"))
                        -- local spawn_pos_y = tonumber(cmeta:get("spawn_pos_y"))
                        -- local spawn_pos_z = tonumber(cmeta:get("spawn_pos_z"))
                        local spawn_pos = {
                                x = 1426,
                                y = 92,
                                z = 1083,
                        }
                        player:set_pos(spawn_pos)
```

pauldpickell:
Hmm, not sure if the UBC schematic sequence is the best example to develop this feature from. Ideally we would have a single schematic to work with. I wonder if it's possible to import another world using the `map.sqlite` file?

pauldpickell:
A possible work around for now might just be to provide the option to set the spawn location for the spawn realm (#67) and define antipodes `pos1` and `pos2` in-game (similar to `Minetest-WorldEdit`) to get the dimensions in order to block off that coordinate space in the realm system. All this without invoking map barriers or the ability to export a schematic. The UBC campus is already built in a singlenode world and generated at `{x=0, y=0, z=0}` so there should be no coordinate conflicts with other realms. 

[...]

PR #93, opened 2022-06-14
Updated "Go Home" button

lukasgolson:
Final feature from https://github.com/ubc-minetest-classroom/Minetest_Classroom/issues/65#issuecomment-1146292044


pauldpickell (commented):
@lukasdragon Can you highlight the change here? I assume its minor, but because the formatting and tabbing was modified, every line is highlighted.

lukasgolson:
> @lukasdragon Can you highlight the change here? I assume its minor, but because the formatting and tabbing was modified, every line is highlighted.

@pauldpickell I've fixed the auto-format/tabbing so that the change is not obscured.

PR #97, opened 2022-06-15
Add missing "end" statements to `mc_worldmanager`

Lowie375:
Fixes a bug that prevents server startup due to missing `end` statements in `mc_worldmanager/commands.lua`


lukasgolson:
I'm looking over why this happened; I'm not sure if the merge conflicts when the changes to the command file were resolved correctly. Looks like a bunch of table entries are no longer formed correctly...

Lowie375:
That seems likely, since `commands["schematic"]` is defined twice in `commands.lua` (L107 and L185)

lukasgolson:
Looks like all commands have been defined twice after the merge into develop;

Lowie375:
Converting to draft since this is a merging issue; naively adding line endings won't solve all the problems uncovered

lukasgolson:
Fixed in #98.

PR #100, opened 2022-06-16
Feature/realm specific privileges

### LINKED ISSUE #36: Realm specific player permissions
From @pauldpickell comment on PR #32 

> This makes me think that we might need to dynamically issue privs (like 'fly') based on the realm. I can foresee teachers not wanting students to be flying around if the task is to dig or look at the ground. Through the Manage Students formspec, teachers can grant/revoke any priv, but we might want to change default privs on-join.

lukasgolson:
This PR solves the first point from issue #36 by introducing realm-specific and universal privileges.

When a player teleports to a realm, their privileges will be calculated and granted by combining realm privileges and "universal privileges".

Realm privileges are privileges that have been assigned to a realm to be given to all players that join that realm for the duration that they remain in that realm. These privileges are managed through the new `realm privs [grant | revoke | list] <realmID> <privilege>` command.  

Universal privileges are those that have previously been granted to a player (managed through the existing privilege management commands: /grant; /grantme; /revoke; and /revokeme). They are active across all realms.


mods/mc_worldmanager/hooks.lua
```diff
@@ -28,8 +30,17 @@ minetest.register_on_respawnplayer(function(player)
     return true
 end)
 
--- When player joins the game, we create their hud
+-- When player joins the game, we make sure all pmeta is set correctly, create their hud, etc.,
 minetest.register_on_joinplayer(function(player, last_login)
+    local pmeta = player:get_meta()
+    if (pmeta:get_int("realm") == nil) then
+        pmeta:set_int("realm", mc_worldManager.GetSpawnRealm().ID)
```
pauldpickell, line 28: Since permissions (realm privileges) are now dependent on realm IDs, and realm IDs are associated to player metadata, what mechanisms prevent a player from somehow spoofing their realm ID or exploiting some other (formspec?) bug that allows them to teleport to a realm and then upgrade their privileges or use privileges that they should not have?

Some attack methods that come to mind:
- formspec injection
- hack client with noclip
- a `teacher` incorrectly grants `all`, `teleport`, `noclip` after the joinplayer callback
- a bug with the player management system noted in point 3 of #36
- breaking the `unbreakable_map_barrier` with a specific tool that has a drop bug
- players inappropriately sharing access codes to classrooms with `fly` privilege and then using some combination of the above

What mechanisms should we implement here to ensure that privileges are not unduly granted? This may be a good opportunity to use a hack client to test vulnerabilities and exploits. 

May also need to reconsider granting `teleport` to any player under universal privileges, including `teacher`s because it might lead to unexpected behavior with the realm teleportation system, which uses the `set_pos()` method.

pauldpickell (dismissed):
I think we need an assessment of what measures we are already implementing and what measures we need to implement to ensure that privileges are not gained maliciously.

mods/mc_worldmanager/hooks.lua
lukasgolson: All very good points that we need to think about and analyze further. I agree that we need to start looking into securing and preventing potential exploits throughout our codebase. Reviewing this PR, it has the potential to be used for privilege escalation if someone had access to the teacher privilege. **Simple fix: require players to hold a privilege to add it to a realm.** ✔️

Thankfully for spoofing realm ID's, It's my understanding that player metadata is stored / updated on the server and is not modifiable by the client. Mechanisms to prevent a player from spoofing their realm ID include that the client cannot call the methods required to teleport to a realm nor update their own (or others) realm ID. 

~~As an added precaution, we could mark the realm ID player metadata key as server-only. This would stop it from being updated on the client making it more tough for students to share knowledge of potentially vulnerable realms (slowing reconnaissance).~~ Edit: `meta:mark_as_private(key)` is only available for nodes. Thankfully, `mark_as_private` might not be necessary as it does not appear that player metadata is usually shared with clients.✔️

**Some general thoughts on the attack methods as outlined:**
- If our mods introduce formspec bugs, we would need to take care of those on a case-by-case basis to prevent arbitrary code execution (e.g., sanitize all client input, never trust what the client is telling us, etc.). Some form of codebase audit would be necessary to verify the current state of our formspecs.
- The realm system is protected from hack clients / noclip by the decision to make realm ID (and subsequent privileges) not based on player location, but on the state of where the server has teleported them. Players might still be able to move to other realms and bother players / teachers, but they won't be granted realm privileges or have other callbacks called. Quick testing for this had me caught by the built-in anti-cheat (https://github.com/minetest/minetest/blob/c510037e9a334b3327a6d6b066203618051e4a09/doc/lua_api.txt#L4883-L4892), but there are likely workarounds that we need to patch.
- We could add a whitelist for the commands a teacher can assign to a realm to prevent arbitrary privilege escalation. ✔️
- The current classroom system implements this feature and has been modified to be an extension to the realms system. I'll add the feature for commands to be applied to all players of a realm, and I'll look into player authentication for realms.
- We can stop players from griefing other classrooms by teleporting them back to the realm they belong if they move out of bounds. ✔️

~~We might want to look into whether it's possible to include CWL authentication as a faculty specific feature to stop players from sharing server information with the internet.~~

To solve this final point, I think we could re-define the teleport command to take into consideration the realms systems.✔️

pauldpickell: My concern about realm ID registered in player meta (which I agree is secured server side) is that currently there is no check (that I am aware of) to test whether the player is physically located in the bounds of that realm. Which opens the possibility of a player "spoofing" their realm ID and using their permissions if they are able to physically move to another realm. Maybe a check can be added somewhere when the player performs an action or at some regular interval (1 min?) to see if player pos is within the range of the current registered realm, and if not, then remove them back to the spawn point for that realm. 

> We might want to look into whether it's possible to include CWL authentication as a faculty specific feature to stop players from sharing server information with the internet.

I know you had mentioned before implementing CWL authentication, but a much simpler and possibly more effective solution might be to set an inbound TCP rule on the server for the port number that the server is listening on to filter out any traffic originating outside of the UBC IP address range. This way, players would need to be on UBC computers or authenticated via UBC VPN to connect (i.e., CWL), which could reduce the risk from malicious clients and players connecting who should not be.

Edit: also maybe adding a time-out function to prevent rapid movement between realms could help reduce grieving. Maybe a 2 min cool-off period between realm hops?

lukasgolson: > My concern about realm ID registered in player meta (which I agree is secured server side) is that currently there is no check (that I am aware of) to test whether the player is physically located in the bounds of that realm. Which opens the possibility of a player "spoofing" their realm ID and using their permissions if they are able to physically move to another realm. Maybe a check can be added somewhere when the player performs an action or at some regular interval (1 min?) to see if player pos is within the range of the current registered realm, and if not, then remove them back to the spawn point for that realm.

Realm ID is not assigned to players unless they are teleported there by the server. This means that if I noclip from realm1 to realm2, I will still be in realm1 according to the server. For my realmID to update, I'd have to get moved there by the `Realm:TeleportPlayer(player)` method found on realm instances. 

I'll implement that check to teleport players back to the realm they belong. It will add the benefit of preventing players from flying where they shouldn't. 

> I know you had mentioned before implementing CWL authentication, but a much simpler and possibly more effective solution might be to set an inbound TCP rule on the server for the port number that the server is listening on to filter out any traffic originating outside of the UBC IP address range. This way, players would need to be on UBC computers or authenticated via UBC VPN to connect (i.e., CWL), which could reduce the risk from malicious clients and players connecting who should not be.

This sounds like the best (and easiest) solution for authentication.

lukasgolson:
Requested changes have been made to resolve the aforementioned concerns. Changes include:
- Restricted which privileges can be assigned to a realm based on those held by the player executing the command. This change is to prevent vertical privilege escalation.
- Added two whitelists to control which privileges are allowed to be granted to a realm. The first one is a mod-configured whitelist, and is by default copied to the world-configured whitelist on world creation (or if the world whitelist is deleted). Both whitelists are called: realmPrivilegeWhitelist.conf

The system was setup this way to allow future configuration by an administrator in-game. The whitelist in the mod folder is write-protected and cannot be saved to at run-time.   

- "void" damage has been added. When a player tries to fly between realms, they will be damaged 4 lives (2 hearts) every 0.5 seconds starting after a short delay (the game checks every 10 seconds to ensure that players are in their realms. This 10 second frequency can be changed in the future based on performance. It should have minimal impact as is, but future optimizations might include iterating the player checks over multiple timesteps). Every 10 seconds, the initial check will increase the amount of damage dealt to players by 4 lives.  This is intended behavior in-case players have somehow boosted their health through glitches / exploits. At max normal health, it will take 2.5 seconds to die.  At max engine allowed health (250 lives),  it will take 20.41 seconds to die in the void. At 5 minutes, it is impossible for players to survive regardless of glitch or exploit. 

The void damage provides players with an effective cue that they should not be leaving the realm while not being overly harsh or annoying for players that accidentally no-clip out of the realm while looking at something. It also solves the issue of players glitching out of the realm with no way to return.  

- Re-defined the teleport command to be compatible with the realms system while maintaining most functionality (will implement teleporting to coordinates later based on the structure `/teleport <realmID> <localX> <localY> <localZ>`)
- Refactored the grantUniversalPriv and revokeUniversalPriv methods so that they can be used by other systems.

lukasgolson:
> Requested changes have been made to resolve the aforementioned concerns. Changes include:
> 
> * Restricted which privileges can be assigned to a realm based on those held by the player executing the command. This change is to prevent vertical privilege escalation.
> * Added two whitelists to control which privileges are allowed to be granted to a realm. The first one is a mod-configured whitelist, and is by default copied to the world-configured whitelist on world creation (or if the world whitelist is deleted). Both whitelists are called: realmPrivilegeWhitelist.conf
> 
> The system was setup this way to allow future configuration by an administrator in-game. The whitelist in the mod folder is write-protected and cannot be saved to at run-time.
> 
> * "void" damage has been added. When a player tries to fly between realms, they will be damaged 4 lives (2 hearts) every 0.5 seconds starting after a short delay (the game checks every 10 seconds to ensure that players are in their realms. This 10 second frequency can be changed in the future based on performance. It should have minimal impact as is, but future optimizations might include iterating the player checks over multiple timesteps).
> 
> The void damage provides players with an effective cue that they should not be leaving the realm while not being overly harsh or annoying for players that accidentally no-clip out of the realm while looking at something. It also solves the issue of players glitching out of the realm with no way to return.
> 
> * Re-defined the teleport command to be compatible with the realms system while maintaining most functionality (will implement teleporting to coordinates later based on the structure `/teleport <realmID> <localX> <localY> <localZ>`)
> * Refactored the grantUniversalPriv and revokeUniversalPriv methods so that they can be used by other systems.

Remaining issues include point 3 of #36 as it requires new infrastructure (such as player tracking in realms) that is not directly related to the realm specific privilege system. Current functionality in player management is functional. I will expand that system in another branch so that this feature branch does not become too large.

mods/mc_worldmanager/commands.lua
```diff
@@ -245,6 +250,72 @@ commands["define"] = {
         Realm:Restore(newRealm)
     end }
 
+commands["privs"] = {
+    func = function(name, params)
+        local operation = tostring(params[1])
+        local realmID = tonumber(params[2])
+        local privilege = tostring(params[3])
+
+        if (operation == nil or operation == "") then
```
Lowie375: `tostring(nil)` on lines [255](https://github.com/ubc-minetest-classroom/Minetest_Classroom/pull/100/files#diff-438bd03cc76fd81ad9448a1f1d70fcb5c719286e2841b32e95fb546e6d172d73R255) and [257](https://github.com/ubc-minetest-classroom/Minetest_Classroom/pull/100/files#diff-438bd03cc76fd81ad9448a1f1d70fcb5c719286e2841b32e95fb546e6d172d73R257) produces `"nil"`, which means that the comparisons on lines [259](https://github.com/ubc-minetest-classroom/Minetest_Classroom/pull/100/files#diff-438bd03cc76fd81ad9448a1f1d70fcb5c719286e2841b32e95fb546e6d172d73R259) and [289](https://github.com/ubc-minetest-classroom/Minetest_Classroom/pull/100/files#diff-438bd03cc76fd81ad9448a1f1d70fcb5c719286e2841b32e95fb546e6d172d73R289) (`"nil" == nil`) will return `false` if `params[1]` or `params[3]` is `nil`

Two patch suggestions are provided below - alternatively, a nil check could be added to lines 255 and 257

```suggestion
        if (operation == "nil" or operation == "") then
```

```suggestion
        if (params[1] == nil or operation == "") then
```

mods/mc_worldmanager/commands.lua
```diff
@@ -245,6 +250,72 @@ commands["define"] = {
...
+                    minetest.chat_send_player(name, "- " .. i)
+                end
+            else
+                minetest.chat_send_player(name, "No realm privileges have been set.")
+            end
+
+            return true, "command executed succesfully"
+        end
+
+        if (privilege == nil or privilege == "") then
```
Lowie375: Same as above
```suggestion
        if (privilege == "nil" or privilege == "") then
```
```suggestion
        if (params[3] == nil or privilege == "") then
```

Lowie375 (changes requested):

mods/mc_worldmanager/commands.lua
```diff
@@ -245,6 +250,72 @@ commands["define"] = {
...
+
+        if (operation == nil or operation == "") then
+            return false, "Incorrect parameter... Missing realm privilege operation. Usage: realm privs [<grant> | <revoke>] <realmID> <privilege>"
+        end
+
+        if (operation == "help") then
+            return true, "Usage: 'realm privs [grant | list | revoke] <realmID> <privilege>'; 'realm privs help' for help."
+        end
+
+        if (realmID == nil or realmID == "") then
```
Lowie375, line 277: `tonumber(nil)` on line [256](https://github.com/ubc-minetest-classroom/Minetest_Classroom/pull/100/files/1fb230731a440999d398ef2da907cc6f43b83610#diff-438bd03cc76fd81ad9448a1f1d70fcb5c719286e2841b32e95fb546e6d172d73R256) produces `nil`, so no patch is needed here

Lowie375 (approved):

pauldpickell (approved):

PR #102, opened 2022-06-18
Polish up `magnify` mod + fully decouple `magnify` from `mc_helpers`

Lowie375:
* Add search bar to plant compendium
* Sort plant compendium alphabetically by common name instead of by reference number
  * Decouples `magnify` and `mc_helpers` due to use of a new sorting algorithm
* Fix item deletion bug when dropping magnifying tool from inside inventory
* Add aliases to magnifying tool for convenience (`magnify:magnifying_glass`, `magnifying_glass`, `magnifying_tool`, `magnify_tool`)
* Add call to `on_punch` callbacks on magnifying tool use for simpler integration into dynamic tutorial framework
* Update returns for `magnify.get_all_registered_species()`


lukasgolson:
Do we want to decouple Magnify from MC_helpers? Or do we want to bring the new algorithm into MC_Helpers and extract the MC_Helpers mod into a new repo?

The Mc_Helpers mod is intended to be a reusable library for all of our mods; making future development faster by eliminating the "reinvention of the wheel". I feel that by removing the dependency, we are defeating the purpose of the library. I'd love to hear any counter arguments though.

pauldpickell:
@lukasdragon I would like to see magnify decoupled completely from MC as a stand alone mod. That said, once we are happy with the state of affairs and create a new repo for magnify, it could have another branch that integrates with MC_helpers, if we want. That branch could be a submodule in MC.

PR #104, opened 2022-06-20
Merge `develop` into `refactor/forestry_tool_management`

Lowie375:
Merges `develop` into `refactor/forestry_tool_management` so that the toolbox and `mc_toolhandler` will be part of the upcoming forestry tool refactor merge into `forestry_tools`. This ensures that the forestry tools will still be available in the player inventory once the refactor is merged, since the refactor removes the existing, outdated inventory management code in the tools

`develop` is being merged instead of `features/tool-menu` since the `features/tool-menu` branch is affected by the issue fixed by #98, and the two branches are very similar in their current state. As a side effect, this will eventually introduce all other changes present in `develop` to `forestry_tools`, but I suspect that will not cause problems since `forestry_tools` is being developed separately from the main MineTest Classroom infrastructure


Lowie375:
Closing for now to prevent accidentally merging any unwanted changes into `develop` again - will reopen from a copy of `develop`

Lowie375:
Re-opening with `develop_merge_copy` as a base

Lowie375:
Superseded by #106

PR #109, opened 2022-06-23
Feature/realm player groups

### LINKED ISSUE #36: Realm specific player permissions

lukasgolson:
This PR adds the following features:
- Added player tracking to the realms system to allow realms integrations to know which players are located within the realm 
- Added callbacks for essential parts of the realms (onRealmCreate, onRealmDestroy, onPlayerJoin, onPlayerLeave). This allows mods to be integrated with the realms system without creating cyclical dependencies
- Simplified the realms API for common tasks (such as getting the realm a player is located, getting a realm from a realm ID)

To solve point 3 from #36: 
- Added realm selection target to the `mc_teacher` mod. This selection target is used by default to help prevent teachers from accidentally targeting all players. 
- Added a "realm" tab to the mc_teacher players menu which displays all players in the current realm
- Fixed the classrooms system to properly work with the realms system again (it had broke in an unknown merge).

### DIFF
--- a/mods/mc_helpers/init.lua
+++ b/mods/mc_helpers/init.lua
@@ -127,3 +127,24 @@ end
 function mc_helpers.trim(s)
     return s:match( "^%s*(.-)%s*$" )
 end
+
+
+function mc_helpers.shallowCopy(table)
+    local copy = {}
+    for k, v in pairs(table) do
+        copy[k] = v
+    end
+    return copy
+end
+
+function mc_helpers.deepCopy(table)
+    local copy = {}
+    for k, v in pairs(table) do
+        if type(v) == "table" then
+            copy[k] = mc_helpers.deepCopy(v)
+        else
+            copy[k] = v
+        end
+    end
+    return copy
+end
--- a/mods/mc_helpers/init.lua
+++ b/mods/mc_helpers/init.lua
@@ -123,3 +123,7 @@ function mc_helpers.pairsByKeys (t, f)
     end
     return iter
 end
+
+function mc_helpers.trim(s)
+    return s:match( "^%s*(.-)%s*$" )
+end
--- a/mods/mc_teacher/gui_dash.lua
+++ b/mods/mc_teacher/gui_dash.lua
@@ -24,7 +24,7 @@ local infos = {
     },
 }
 local tool_name = "mc_teacher:controller"
-local priv_table = {teacher = true}
+local priv_table = { teacher = true }
 
 local function get_group(context)
     if context and context.groupname then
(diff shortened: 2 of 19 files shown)


mods/mc_helpers/init.lua
```diff
@@ -123,3 +123,28 @@ function mc_helpers.pairsByKeys (t, f)
     end
     return iter
 end
+
```
Lowie375, line 126: MineTest provides some [helper functions](https://minetest.gitlab.io/minetest/helper-functions) which seem to be very similar in functionality to the newly-added helper functions:

- `mc_helpers.trim(s)` and `string:trim()`
- `mc_helpers.deepCopy(table)` and `table.copy(table)`

That being said, it may still be good to keep these implementations so that we have helper functions which have more predictable behaviours

mods/mc_teacher/gui_dash.lua
```diff
@@ -206,9 +213,11 @@ local function get_player_list_formspec(player, context)
     do
         fs[#fs + 1] = "tabheader[0,0;6.7,0.8;group;"
         fs[#fs + 1] = FS "All"
+
         local selected_group_idx = 1
         local i = 2
-        for name, group in pairs(minetest_classroom.get_all_groups()) do
+
+        for name, group in pairs(minetest_classroom.get_all_groups(player)) do
```
Lowie375: `minetest_classroom.get_all_groups()` as defined on [line 77 of `api.lua`](https://github.com/ubc-minetest-classroom/Minetest_Classroom/pull/109/files?diff=unified&w=1#diff-b84c94f3b3843d24b408d0d2938cfce2cefbf9892cd5cfbd3db3e2c8b04e5c6cR77) does not take any arguments

mods/mc_worldmanager/commands.lua
```diff
@@ -29,13 +23,23 @@ minetest.register_chatcommand("localPos", {
 
 commands["new"] = {
     func = function(name, params)
-        local realmName = params[2]
+        local realmName = tostring(params[1])
         if (realmName == "" or realmName == nil) then
```
Lowie375: [Same issue with unintended `"nil" == nil` comparison](https://github.com/ubc-minetest-classroom/Minetest_Classroom/pull/100#discussion_r910437827) as in #100, due to `tostring(nil)` producing `"nil"`
```suggestion
        if (realmName == "" or realmName == "nil") then
```
```suggestion
        if (realmName == "" or realmName == params[1]) then
```

mods/mc_worldmanager/commands.lua
```diff
@@ -29,13 +23,23 @@ minetest.register_chatcommand("localPos", {
...
             realmName = "Unnamed Realm"
         end
-        local size = params[3]
-        local sizeY = params[4]
-        local newRealm = Realm:New(realmName, { x = size, y = sizeY, z = size })
+        local sizeX = tonumber(params[2])
```
Lowie375, line 30: `tonumber(nil)` produces `nil`, so the issue on lines [26-27](https://github.com/ubc-minetest-classroom/Minetest_Classroom/pull/109/files?diff=unified&w=1#diff-438bd03cc76fd81ad9448a1f1d70fcb5c719286e2841b32e95fb546e6d172d73R26-R27) does not apply here

Lowie375 (dismissed):

mods/mc_helpers/init.lua
lukasgolson: These versions have been implemented because the minetest versions do not appear to be JIT compiled in lua JIT versions of Minetest, increasing context switching and reducing performance. Although it's a micro-optimization, the functions were simple enough to implement that it offset the added complexity.

Lowie375: Noted; will make a note to refactor calls to `string:trim()` and `table.copy()` in some other mods once this branch has been merged

mods/mc_teacher/gui_dash.lua
lukasgolson: Great catch; I'm not sure why I changed this line to add the player param. Will revert this change asap.

PR #117, opened 2022-07-05
Features/worldgen

lukasgolson:
This PR implements the Lua scriptable world generation framework/pipeline/[insert keyword here] for realms. The framework is heavily simplified and built from initial testing and experimentation with world generation. I will add documentation to everything once more pressing issues have been completed.

The pipeline enforces a 3-step process used internally by games such as Minecraft and Minetest. First, a heightmap is generated in stone, water, and air. Surface details such as lakes, lava ponds, etc., Secondly, surface layers (e.g., dirt, grass) replace stone and details are added. Finally, trees and schematics are placed.  

It should be noted that when generating a world, step 1 is independent but steps 2 and 3 are linked.

With this PR, 2 height map generators and 3 decorators have been created for R&D testing. They are all viable as are but have lots of room for improvement. The two height map generators take less than a second to generate large (up to 2,000 X 2,000) worlds, while the decorators [v1, v2] are slower (but still faster than the vanilla world generator). 

The biomegen decorator is about the same speed or slightly slower than the vanilla decorator.

The V1 height map generator uses simple perlin noise to create plains / flatlands like terrain.

The V2 height map generator builds on V1 to add mountainous hills, and more interesting terrain features. It also introduces cliffs under rare circumstances.

The V1 world decorator adds simple grass and sand over the terrain in appropriate locations.

The V2 world decorator builds on V1 to add L-system trees. I want to expand on this version once we have the tree system in place.

The biomegen world decorator interfaces with a modified version of the biomegen mod to decorate the terrain with vanilla-registered biomes and features from the base game / other mods.


mods/mc_helpers/PointTable.lua
```diff
@@ -28,4 +28,33 @@ function ptable.delete(table, coords)
...
+function ptable.get2D(table, coords)
+    return table[coords.x] and table[coords.x][coords.y]
+end
+
+
+function ptable.delete2Ds(table, coords)
+    if ptable.get(table, coords) == nil then
+        return
+    end
+    local tx = table[coords.x]
```
Lowie375, line 52: Consider adding a `nil` check to `tx` to prevent an error from being thrown if the user attempts to delete a position with an `x` coordinate that has not been defined

lukasgolson: I thought about that, but the proper course of action would be to throw an error if we were to be provided with a malformed positional vector. With this data structure, there will always be an x coordinate: https://minetest.gitlab.io/minetest/spatial-vectors/

There is the philosophy that all parameters should be checked. I agree with this. The correct alternative to what we currently have would be a nil check and throw using another error message.

Lowie375 (approved):

Issue #89, opened 2022-06-14
A framework for describing different classes of realms


lukasgolson:
From @pauldpickell on Slack: 

> A framework for describing different classes of realms, even if just conceptually, besides just instanced/spawn/other. Something like `classrooms `(created only by teachers that authorized players and any teacher can join), `instanced `(created by a player and only visible/joinable by that player and any teacher), and `spawn `(created only by teacher at game initialization and joinable/visible to all players). This way, we have a way to sort privileges, access, and perhaps other features based on the purpose for the realm.

lukasgolson:
Started working on realm categories on branch: https://github.com/ubc-minetest-classroom/Minetest_Classroom/tree/feature/realmCategories

Development on this feature is stalled until dependencies in #100 and #109 have been merged into develop.

PR #125, opened 2022-07-15
Feature/realm categories

lukasgolson:
This PR adds the realm categories feature as described in #89, solidifies instanced and temporary realms, and adds a teleporter block for easier player movement. 

For an example of how this teleport block might work with tutorial realms, execute the following commands `/realm blocks teleporter 1 true true punchablock punchablockschematic` to receive a block that will teleport you to a temporary, instanced realm containing the punchablock schematic.


mods/mc_worldmanager/realm/realmCategory.lua
```diff
@@ -0,0 +1,135 @@
...
+
+        if (realm:get_data("students") == nil) then
+            realm:set_data("students", {})
+        end
+
+        if (realm:get_data("owner") == nil) then
+            realm:set_data("owner", {})
+        end
+
+        if (realm:get_data("students")[player:get_player_name()] ~= nil) then
```
pauldpickell, line 82: How are teachers treated? Need to consider the case that teachers should be able to join any realm (e.g., needing to help a student or grade an assignment/project), so perhaps they need to be added as owners?

mods/mc_worldmanager/realm/realmPlayerManagement.lua
```diff
@@ -1,8 +1,16 @@
...
 ---TeleportPlayer
 ---Teleports a player to this realm.
----@param player objectRef
+---@param player
 function Realm:TeleportPlayer(player)
+    local realmCategory = self:getCategory()
+
+    local joinable, reason = realmCategory.joinable(self, player)
+
+    if (not joinable and not minetest.check_player_privs(player, { teacher = true })) then
```
pauldpickell, line 10: Okay, I see this now. So if I understand correctly, the teacher has to be added as either a student or an owner of the realm?

mods/mc_worldmanager/teleporterNode.lua
```diff
@@ -0,0 +1,190 @@
...
+                else
+                    realmObject = Realm:New(realmName, { x = 80, y = 80, z = 80 })
+                    realmObject:CreateGround()
+                    realmObject:CreateBarriersFast()
+                end
+
+                meta:set_int("realm", realmObject.ID)
+                minetest.swap_node(pos, { name = "mc_worldmanager:teleporter", param2 = math.ceil(math.sin(realmObject.ID) * 255) })
+            else
+                minetest.chat_send_player(clicker:get_player_name(), "This teleporter is broken. Please contact an instructor for assistance.")
```
pauldpickell, line 89: Might also be useful to indicate that the student can use the reporting function in the student notebook to report this issue.

pauldpickell (changes requested):
Minor suggestions and some questions.

mods/mc_worldmanager/realm/realmCategory.lua
lukasgolson: Users with the teachers privilege are assumed to be excluded from this check to make implementation details easier for any add-on mod. 

E.g., add-on mods should respect that teacher privileges override the check without creating a dependency.

See the teleportPlayer function in the realm player management file for an example of the assumption I'm making. 

I'll see if I can move this implementation from calling methods into this class.

mods/mc_worldmanager/realm/realmPlayerManagement.lua
lukasgolson: See my above response to previous comment for context.

While this is true, it's not necessary. Players with the teacher privilege are not checked with this system under the assumption that calling functions will handle these players separately.

mods/mc_worldmanager/teleporterNode.lua
lukasgolson: Good idea, I'll look into this.

PR #129, opened 2022-07-19
Tutorial recording system overhaul

### LINKED ISSUE #145: Feature: Dynamic tutorial framework improvements
### Planned for [#129](../pull/129)
- [x] Allow multiple players to record/access tutorials simultaneously
- [x] Redesign tutorial saving formspec
- [x] Add event groups (for when the order of actions that need to be done at a certain stage in the tutorial does not matter)
  - [x] Make event groups functional in tutorials
- [x] Allow multiple items/privileges to be awarded on completion
- [x] Allow the event sequence to be modified after recording
  - [x] Add new events
  - [x] Edit existing events
  - [x] Reorder events
  - [x] Duplicate events
- [x] Add tutorial dependencies/dependents
  - [x] Display dependencies/dependents in tutorial book
  - [x] Highlight tutorials that are locked
- [x] Add confirmation for tutorial book delete button
- [x] Add confirmation button + warnings for potentially dangerous privs (all, noclip, privs, etc.)

### Upcoming
- [ ] Display next actions during an active tutorial
- [ ] Add static messages in HUD, formspec, or chat
- [ ] Implement additional functionality
  - [ ] Search bars for tutorials/rewards
  - [ ] Reward quantity input
- [ ] Add new event types
  - [ ] Crafting
  - [ ] Static events
    - [ ] Popups
    - [ ] Chat messages
    - [ ] HUD events
  - [ ] Inventory actions
    - [ ] `put`
    - [ ] `take`
    - [ ] `move`
  - [ ] Relative position (realm-specific)
- [ ] Add option to make tutorials realm-specific
  - [ ] Allow an instanced realm to be created when a tutorial is started
    - [ ] Remove instanced realm on: player leave, tutorial stop, tutorial completion(?)
  - [ ] Allow tutorial-granted privs to be realm-specific or global
    - [ ] Realm-specific granting requires recorder to have realm ownership
    - [ ] Global granting requires requires recorder to have the `privs` privilege (or a new admin privilege?)
- [x] Add tutorial related commands (`/tutorial`)
  - [ ] `/tutorial start`
  - [ ] `/tutorial stop`
  - [ ] `/tutorial next` - tells the player what they must do next to progress
  - [x] `/tutorial list` (replacing `/listTutorials`)
- [ ] Merge old tutorial system (`mc_tutorial_framework`) into dynamic system (`mc_tutorial`)
  - [ ] Migrate `mc_tutorial_framework` tutorials to `mc_tutorial` format
  - [ ] Populate tutorial book with `mc_tutorial_framework` tutorials on world creation

Lowie375:
This PR is part 1 of the tutorial recording system overhaul, which contains many of the core features of the overhaul. This has been opened in its current state so that user tests of this system can be done on a build from the `develop` branch for simplicity.

- Redesign the tutorial recording menu so that it is easier to use
- Allow multiple items and privileges to be given on tutorial completion
- Allow tutorial action grouping
  - Players can complete actions within a group in any order
  - The tutorial will progress once all actions in the group have been completed
- Add support for tutorial dependencies/dependents

Please see #145 for more detailed information about what this PR includes and what is on the roadmap for part 2 of the tutorial recording system overhaul.


pauldpickell:
This formspec looks so much better! Just a thought, might we add an `item_image` element above or below the Available Rewards list that shows the highlighted element in the list? The technical names might not mean much to the lay user. Maybe we need icons for the privs?

Lowie375:
> Just a thought, might we add an `item_image` element above or below the Available Rewards list that shows the highlighted element in the list?

Yes, that's a great idea! Will work on adding that in

> Maybe we need icons for the privs?

This might be helpful too, since some privs aren't very simple to understand from name alone. Will also see if any privs have descriptions that we could use as well
UPDATE: Privileges do have descriptions, will look into integrating a way to view that as well

mods/mc_tutorial/init.lua
```diff
@@ -0,0 +1,1708 @@
...
+                -- Send to mod storage
+                local tutorials = mc_tutorial.tutorials:to_table()
+                if not tutorials or not next(tutorials.fields) then
+                    mc_tutorial.tutorials:set_int("next_id", 2)
+                    mc_tutorial.tutorials:set_string("1", minetest.serialize(recorded_tutorial))
+                else
+                    local id = mc_tutorial.record.edit[pname] or mc_tutorial.tutorials:get("next_id") or 1
+                    
+                    -- Update dependencies/dependents in other tutorials
+                    for k,v in pairs(mc_tutorial.record.temp[pname].depend_update.dep_cy) do
```
pauldpickell, line 1959: This assigns a boolean to `k`, but the next line expects a string for `get()`.
Can reproduce by creating any tutorial and then attempting to set the dependency and it crashes with this output:
```
2022-08-18 21:02:27: ERROR[Main]: ServerError: AsyncErr: Lua: Runtime error from mod 'mc_tutorial' in callback on_playerReceiveFields(): ...om-features-tutorial_recording\mods\mc_tutorial\init.lua:1439: bad argument #1 to 'get' (string expected, got boolean)
2022-08-18 21:02:27: ERROR[Main]: stack traceback:
2022-08-18 21:02:27: ERROR[Main]: 	[C]: in function 'get'
2022-08-18 21:02:27: ERROR[Main]: 	...om-features-tutorial_recording\mods\mc_tutorial\init.lua:1439: in function <...om-features-tutorial_recording\mods\mc_tutorial\init.lua:971>
2022-08-18 21:02:27: ERROR[Main]: 	...st\minetest-5.5.1-win64\bin\..\builtin\game\register.lua:429: in function <...st\minetest-5.5.1-win64\bin\..\builtin\game\register.lua:415>
```

mods/mc_tutorial/init.lua
```diff
@@ -0,0 +1,1708 @@
...
+    for _,priv in pairs(mc_tutorial.fetch_setting_table(name) or {}) do
+        output[priv] = true
+    end
+    return output
+end
+
+mc_tutorial.player_priv_table = mc_tutorial.fetch_setting_key_table("player_priv_table") or {interact = true}
+mc_tutorial.recorder_priv_table = mc_tutorial.fetch_setting_key_table("recorder_priv_table") or {teacher = true, interact = true}
+mc_tutorial.check_interval = math.max(tonumber(mc_tutorial.fetch_setting("check_interval")), 0.1) or 1
+mc_tutorial.check_dir_tolerance = mc_tutorial.fetch_setting("check_dir_tolerance") or 0.01745
```
pauldpickell: This `check_dir_tolerance` still seems to sensitive with the global step to pick up the action. I wonder if, in addition to the look direction listener, we could simply do a ray cast and instead obtain the node that the player is looking at (within reason), then just check if the player is looking at that node? I suspect that is mainly how this functionality would be used.

mods/mc_tutorial/init.lua
```diff
@@ -0,0 +1,1708 @@
...
+                else
+                    table.insert(rewards, {col = "#FFCCFF", s = minetest.formspec_escape("[P] ")..priv})
+                end
+            end
+
+            local item_map = {
+                ["tool"] = {col = "#CCFFFF", s_pre = minetest.formspec_escape("[T] ")},
+                ["node"] = {col = "#CCFFCC", s_pre = minetest.formspec_escape("[N] ")},
+            }
+            for item,def in pairs(minetest.registered_items) do
```
pauldpickell, line 473: For our purposes, should we also undertake a check against the `mc_toolhandler` state to ensure that we are not granting tools that players should not have more than one copy of or do not have the privileges for?

mods/mc_tutorial/init.lua
```diff
@@ -0,0 +1,1708 @@
...
+                local item_trim = mc_helpers.trim(item)
+                if item_trim ~= "" then
+                    local raw_col_field = item_map[def.type] or {col = "#FFFFCC", s_pre = minetest.formspec_escape("[I] ")}
+                    if mc_helpers.tableHas(temp.on_completion.items, item) then
+                        table.insert(selected_rewards, {col = raw_col_field.col, s = raw_col_field.s_pre..item_trim})
```
pauldpickell: It would be helpful to check and then indicate in the formspec if there are any special privileges needed to use the tool, so that the recording player is aware that they might also need to grant that privilege when the player completes the tutorial. 

Additionally, I propose restricting privileges available to grant to only those that are available to the player. One exploit I can foresee is that anyone with `teacher` who can access this tool can then grant server and moderation-level privileges that they do not (and should not) have access to.

We might also want to add an additional warning for specific privileges that relate to world-manipulation, moderation, administration, and also these specific privs: `noclip`, `teleport`, `bring`, `settime`, and `teacher`. Not sure if this warning ought to be simply red text somewhere or a popup or other interactive element that needs to be confirmed by the recording player before the tutorial is saved.

mods/mc_tutorial/functions.lua
```diff
@@ -0,0 +1,321 @@
...
+            end
+
+            -- TODO
+            local player_privs = minetest.get_player_privs(pname)
+            -- add granted privs to universal priv table
+            if minetest.get_modpath("mc_worldmanager") then
+                mc_worldManager.grantUniversalPriv(player, pdata.active.on_completion.privs)
+            end
+            -- grant privs
+            for _,priv in pairs(pdata.active.on_completion.privs) do
```
pauldpickell, line 349: This does not work for me. When I set a privilege as a reward, it is not granted when I complete the tutorial.

mods/mc_tutorial/init.lua
```diff
@@ -0,0 +1,1708 @@
...
+        "button_exit[3.7,1.4;3,0.8;wieldeditem;Wielded Item]",
+        "button_exit[7,1.4;3,0.8;playercontrol;Pressed Key]",
+        "button_exit[3.7,3.7;3,0.8;exit;Nevermind!]"
+    }
+    local pname = player:get_player_name()
+	minetest.show_formspec(pname, "mc_tutorial:record_options_fs", table.concat(record_options_fs, ""))
+	return true
+end
+
+function mc_tutorial.show_event_popop_fs(player, is_edit)
```
pauldpickell: This is awesome! However, this context can only be accessed once a tutorial has been recorded and the player has entered the edit mode, which requires that the recording player records one of the limited actions. Can we make this available also as a button `Enter Manually` on the `on_rightclick` formspec menu for the recording tool? It would be nice to be able to access this context when initially recording the tutorial, because in some cases it is easier and more intuitive to record certain actions like keystrokes and wielded items. For example, I could record an entire movement tutorial just based on keystrokes without ever needing to use the interactive "listener" mode.

mods/mc_tutorial/init.lua
```diff
@@ -0,0 +1,1708 @@
...
+                            end,
+                        },
+                        [mc_tutorial.DIR.VECTOR] = {
+                            name = "Spatial vector",
+                            get = function()
+                                return table.concat({
+                                    "label[0.4,3.2;X =]",
+                                    "label[0.4,4.1;Y =]",
+                                    "label[0.4,5.0;Z =]",
+                                    "field[1,2.8;6.4,0.8;dir_x;;", context.epop.fields.dir and context.epop.fields.dir.x or "", "]",
```
pauldpickell: Okay, when I enter `[x,y,z]` coordinates here and save the tutorial then go back and edit this action, the values have been replaced with different values. I expected to enter world coordinates here, but it looks like they were converted to radians? Regardless, the `look_dir` listener is not registering the completed action as expected. Maybe it is too sensitive or maybe I am misunderstanding how to use this?

mods/mc_tutorial/init.lua
```diff
@@ -0,0 +1,1708 @@
...
+                    }
+                end,
+                expanded_elem = function()
+                    return {
+                        "button[0.4,2.9;3.4,0.8;key_add;Add selected]",
+                        "button[4,2.9;3.4,0.8;key_delete;Remove selected]",
+                    }
+                end,
+            },
+            [mc_tutorial.ACTION.LOOK_YAW] = {
```
pauldpickell, line 891: Remove the sidebar for this action type. Currently shows registered items, which is unrelated.

mods/mc_tutorial/init.lua
```diff
@@ -0,0 +1,1708 @@
...
+                name = "Look in horizontal direction (LOOK_YAW)",
+                fs_elem = function()
+                    return {
+                        "field[0.4,2.3;7,0.8;yaw;Yaw (horizontal degrees);", context.epop.fields.yaw or "", "]",
+                        "field_close_on_enter[yaw;false]",
+                    }
+                end,
+            },
+            [mc_tutorial.ACTION.LOOK_PITCH] = {
```
pauldpickell, line 906: Remove the sidebar for this action type. Currently shows registered items, which is unrelated.

mods/mc_tutorial/init.lua
```diff
@@ -0,0 +1,1708 @@
...
+                name = "Look in vertical direction (LOOK_PITCH)",
+                fs_elem = function()
+                    return {
+                        "field[0.4,2.3;7,0.8;pitch;Pitch (vertical degrees);", context.epop.fields.pitch or "", "]",
+                        "field_close_on_enter[pitch;false]",
+                    }
+                end,
+            },
+            [mc_tutorial.ACTION.LOOK_DIR] = {
```
pauldpickell, line 921: Remove the sidebar for this action type. Currently shows registered items, which is unrelated.

mods/mc_tutorial/init.lua
```diff
@@ -0,0 +1,1708 @@
...
+
+                    context.epop.d_input_type = context.epop.d_input_type or 1
+                    return {
+                        "label[0.4,2.2;Input type]",
+                        "dropdown[2.1,1.9;5.3,0.6;dir_type;", table.concat(input_types, ",") , ";", context.epop.d_input_type, ";true]",
+                        input_map[context.epop.d_input_type] and input_map[context.epop.d_input_type].get() or "",
+                    }
+                end,
+            },
+            [mc_tutorial.ACTION.POS_ABS] = {
```
pauldpickell, line 976: Remove the sidebar for this action type. Currently shows registered items, which is unrelated.

mods/mc_tutorial/init.lua
```diff
@@ -0,0 +1,1708 @@
...
+                elseif mc_helpers.tableHas(pdata.completed, id) then
+                    col = "#71EBA8"
+                end
+                table.insert(titles, (col or "")..tutorial_info.title)
+            end
+        end
+        context.tutorial_selected = context.tutorial_selected or 1
+        local selected_info = minetest.deserialize(tutorials.fields[tostring(context.tutorial_i_to_id[context.tutorial_selected])] or minetest.serialize(nil))
+        
+        fs = {
```
pauldpickell, line 1163: We ought to modify actions that are available based on the tutorial state. There are two currently three states that are relevant for the tutorial list:

- `In progress` = the player is currently attempting the tutorial
- `Completed` = the player has already done the tutorial
- `Incomplete` = the player has never completed all actions of the tutorial or never attempted  the tutorial

I like that completed tutorials are indicated with green text. However, when the player is `In progress` with any tutorial, then the `edit` and `delete` fields should not be visible. I was able to delete a tutorial that I was attempting with the consequence that a listener was still active. Therefore, if a tutorial is in an `In progress` state, then we need a button to stop the tutorial and return it to an editable/deleteable state to ensure all listeners are also turned off. I have only been testing in a singleplayer server environment, but it would probably be good to ensure that a tutorial is returned to the `Incomplete` state and any listeners are turned off when a player leaves the game.

UX/UI: Just flagging now that we might want to consider how best to display these states to the player. I think this is a broader challenge that could also solve issues with visualizing the dependencies/dependents somehow. But the current way of organizing tutorials in a single list with only colours distinguishing them is not particularly intuitive. @mandyhuy Can you take a look at this from a UX/UI perspective?

mods/mc_tutorial/init.lua
```diff
@@ -0,0 +1,1708 @@
...
+                    })
+                    table.insert(context.events, context.selected_event + 1, "#CCFFFF} END GROUP "..group)
+                    table.insert(mc_tutorial.record.temp[pname].sequence, context.selected_event, {
+                        action = mc_tutorial.ACTION.GROUP,
+                        g_type = mc_tutorial.GROUP.START,
+                        g_id = group,
+                    })
+                    table.insert(context.events, context.selected_event, "#CCFFFFGROUP "..group.. " {")
+
+                    mc_tutorial.record.temp[pname].next_group = group + 1
```
pauldpickell, line 1781: This leads to sequential group numbering even after the earlier groups are deleted. For example, I can create and delete 3 groups, then create another group and the index starts at GROUP 4 instead of 1.

pauldpickell (changes requested):
With great power comes great responsibility. I think we have a decision point around whether this is a standalone mod that should work in any other minetest context or if it should be core to mc.

One critical issue is that privileges are currently a possible reward, but this conflicts with how `mc_worldmanager` constrains player context. If we grant a privilege, is it supposed to apply across all realm categories or only specific realm categories? Is it temporary or only for the instanced realm for the tutorial? In order to access specific realms, will it be necessary to complete training first in a tutorial (i.e., realm-tutorial dependency)? If we have a tutorial that rewards `noclip`, then should the `Realm.whitelistedPrivs` be updated for the player so that whenever they create an instanced realm they can `noclip`? The answer to this last question is decidedly "no" in our mc context because we do not want players leaving the bounds of the realms, but maybe the answer is "yes" for other privs?

@lukasdragon can these questions be easily addressed with the world/player manager framework?

Another mc context issue with the recording system is that it does not consider areas or protection. So the recording player could conceivably record or create a tutorial that has actions that another player could never complete. For example, an area might be protected so that if I try to dig, I only end up punching nodes, but I could manually add a dig action. Maybe we need more validation of action logic?

mods/mc_tutorial/init.lua
pauldpickell: Edit: we should never allow `noclip` to be granted in mc.

mods/mc_tutorial/init.lua
Lowie375: Hm, interesting. I thought I set things up so that the booleans would be in the values (`v`) and not the keys (`k`),  will check and see what's going on

mods/mc_tutorial/init.lua
Lowie375: That might actually work better as a separate action (`LOOK_NODE`), since using a node for comparison would mean that the player could look at it from any direction to complete the action.

mods/mc_tutorial/init.lua
Lowie375: Good idea! Will see what I can do to make this possible

mods/mc_tutorial/init.lua
Lowie375: This is a spatial vector input, not a coordinate input.
Since it could be confusing to some, I could remove this input method altogether and restrict this to a yaw/pitch input, since that is simpler.

mods/mc_tutorial/init.lua
Lowie375: Noted; this is being done to ensure that no groups have internal IDs that overlap. Can look into changing the way groups are numbered so that group IDs automatically update as groups are removed

mods/mc_tutorial/init.lua
pauldpickell: Doh! This makes so much more sense. I think we can leave it in, but tool tips might be useful or a short description in the sidecar to the right. Maybe re-order the input options so that yaw/pitch is the default?

mods/mc_tutorial/init.lua
Lowie375: Will add this check when the tutorial is giving tools so that rewards can still be added to tutorials and given if `mc_toolhandler` is later removed or deactivated, but they aren't given if `mc_toolhandler` is actively managing them

mods/mc_tutorial/init.lua
Lowie375: Was unable to replicate this, suspect that it might be caused by an issue with the ID extraction helper function or old format tutorials and storage conversion.

Lowie375: Have added some additional checks that may prevent this from throwing an error, could you let me know what happens when you perform that sequence of actions now?
Additionally, could you let me know if you see ID numbers before the tutorials in the "Available tutorials" list in the "Dependencies" tab of the recording menu?

lukasgolson:
> One critical issue is that privileges are currently a possible reward, but this conflicts with how `mc_worldmanager` constrains player context. If we grant a privilege, is it supposed to apply across all realm categories or only specific realm categories? Is it temporary or only for the instanced realm for the tutorial? In order to access specific realms, will it be necessary to complete training first in a tutorial (i.e., realm-tutorial dependency)? If we have a tutorial that rewards `noclip`, then should the `Realm.whitelistedPrivs` be updated for the player so that whenever they create an instanced realm they can `noclip`? The answer to this last question is decidedly "no" in our mc context because we do not want players leaving the bounds of the realms, but maybe the answer is "yes" for other privs?
> 
> @lukasdragon can these questions be easily addressed with the world/player manager framework?

I think this is where the distinction between universal and realm-based privileges come into play and are a features which we can take advantage of. Whether we want to grant as a reward for completing quests, universal privileges that apply across all realms or more localized realm-specific privileges, either can be easily done. The framework does not necessarily limit nor dictate how privileges are used -- only the way that they are applied has been made more precise.

As for granting Noclip, I don't foresee that it could be an issue if it's granted to players. Those who try to retaliate against our educational regime and attempt escape, will promptly be eliminated and respawned within the confines of their educational cell. I could see that noclip is probably appropriate for some contexts (building scenarios) but not all (e.g., measurement scenarios). Further thought and discussion should be made around this, but I take the rushed stance/opinion that noclip should remain an option as we have mechanisms to counter its exploitation.

mods/mc_tutorial/init.lua
lukasgolson: > 

See my previous comment for why I disagree with never allowing `noclip `to be granted (within reason).

mods/mc_tutorial/init.lua
Lowie375: Updated to show information about the various input types for this mode

mods/mc_tutorial/init.lua
Lowie375: Have not removed the sidebars for the `LOOK_PITCH`, `LOOK_YAW` and `POS_ABS` events since that would result in formspec size changes without clicking on the expand/collapse button, which may disrupt the player when using the formspec.
Have instead added a "no related information" message to sidebar, replacing the item list.

mods/mc_tutorial/init.lua
Lowie375: > It would be helpful to check and then indicate in the formspec if there are any special privileges needed to use the tool, so that the recording player is aware that they might also need to grant that privilege when the player completes the tutorial.

This may be easier said than done, since there isn't a set way to get the privileges needed to use a tool. However, I could add the privileges `mc_toolhandler` requires for the tool to be given (for tools it manages), along with a warning that the tool is being managed by `mc_toolhandler`

mods/mc_tutorial/init.lua
Lowie375: I'll implement this along with the other new tutorial actions as part of the upcoming second part of this PR. Right now, it's more important to get known bugs fixed and in time for the upcoming user tests.

mods/mc_tutorial/init.lua
Lowie375: Have marked the active tutorial in blue and added more checks to prevent active tutorials from being edited and/or deleted.

mods/mc_tutorial/init.lua
Lowie375: Existing group IDs are now automatically re-indexed when groups are deleted

mods/mc_tutorial/init.lua
pauldpickell: Hm, you are right. This might be a nice request to the core developers, something like `minetest.get_tool_privs`. For now, we can at least manage the teacher controller.

mods/mc_tutorial/init.lua
pauldpickell: These sidecars could be nice to eventually add more detailed instructions.

lukasgolson (commented):
Looks good to me; I don't see any issues after a simple review.

mods/mc_tutorial/init.lua
Lowie375: Have added warning messages to the descriptions of dangerous privileges + an additional confirmation popup for especially dangerous ones (`teacher`, `ban`, `privs`, `password`, `server`, etc.)

Will work on restricting grantable privs to those the recorder has. I think having the `privs` privilege should override this, since `privs` allows the user to grant any privileges (and likewise with `basic_privs`, although for a smaller subset of privileges)  
Would it be worthwhile to also restrict privilege reward removal in the same way? (i.e. if a previously recorded tutorial grants `priv`, editors must also have `priv` in order to remove it from the reward list)

pauldpickell (approved):

PR #130, opened 2022-07-20
Feature/world schematics

lukasgolson:
This PR adds the following features:
- Exporting a realm schematic now also registers it with the realm schematic manager so that it can be used to create new realms right away, (improving on the old workflow where we had to manually copy the schematic into a new mod and register in init).
- When the server starts, any valid realm schematic config files in the `world/realmSchematics` are now automatically registered with the realm schematic manager..
- A crash-causing bug with realm commands and forgetting to supply realmID when executing commands has been fixed.


Lowie375 (approved):
Looks good!
Will also note that the file path separator used is `\` - not an issue, just useful to note for testing purposes so that we can figure out which separators work best cross-platform

lukasgolson:
> 

File path separators are always a bit finicky... I think we need to look into it further. I have no idea if I'm doing it correctly. We'll have to do lots of testing. I'll do some searching online to see if there is a proper way to do it.

pauldpickell:
If I understand correctly, any schematics saved in-game will only be available when the server restarts? Can we expose the file path to make them available immediately? Just thinking it may not be intuitive to a first time user that is not aware of needing to restart the server before they can place the schematic again.

lukasgolson:
> If I understand correctly, any schematics saved in-game will only be available when the server restarts? Can we expose the file path to make them available immediately? Just thinking it may not be intuitive to a first time user that is not aware of needing to restart the server before they can place the schematic again.

They're available right away. When the schematic is saved, it's registered for immediate use. Sorry about the confusion.

PR #134, opened 2022-07-21
Made players immortal by default at game start; removes immortality w…

lukasgolson:
A quick simple fix that makes the previous damage-based player management system effective again. 

This fix removes player invulnerability while players are outside of a realm. When players enter back into the realm, they are made invulnerable again.

### DIFF
--- a/mods/mc_worldmanager/hooks.lua
+++ b/mods/mc_worldmanager/hooks.lua
@@ -2,6 +2,11 @@ local function makePlayerImmortal(player)
     local groups = player:get_armor_groups() or {}
     groups["immortal"] = 1
     player:set_armor_groups(groups)
+    
+    local prop_table = player:get_properties()
+    if prop_table.hp_max then
+        player:set_hp(prop_table.hp_max, "immortality")
+    end
 end
 
 local function makePlayerMortal(player)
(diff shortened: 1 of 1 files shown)


mods/mc_worldmanager/hooks.lua
```diff
@@ -1,3 +1,16 @@
+local function makePlayerImmortal(player)
+    local groups = player:get_armor_groups() or {}
+    groups["immortal"] = 1
+    player:set_armor_groups(groups)
```
Lowie375, line 4: Making a player immortal doesn't reset their HP, so if they took damage, their health bar will still show up after becoming immortal again
Healing the player could be a good way to show the player that they are immortal again, since the health bar is hidden if a player is immortal and at full health

```suggestion
    player:set_armor_groups(groups)
    
    local prop_table = player:get_properties()
    if prop_table.hp_max then
        player:set_hp(prop_table.hp_max, "immortality")
    end
```

Lowie375 (approved):
Looks good!
Have also added one suggestion to `makePlayerImmortal` which may improve quality of life a bit

mods/mc_worldmanager/hooks.lua
lukasgolson: I like this suggestion, good call!

PR #135, opened 2022-07-22
Feature/schematic offset

### LINKED ISSUE #144: Bug: Generating a realm from a schematic crashes the game
**Describe the bug**
Server crash occurred on an instance running from commit [`3dfa4a3`](https://github.com/ubc-minetest-classroom/Minetest_Classroom/commit/3dfa4a3fde92d8d905d4c7f029061ac2ef463ca5) on `feature/schematicOffset` when trying to generate a realm from the `MKRF512_all` schematic using commands:
```
AsyncErr: Lua: Runtime error from mod '' in callback on_chat_message(): ...om\mods\mc_worldmanager/realm/realmSchematicSaveLoad.lua:98: attempt to call method 'LocalToWorldPosition' (a nil value)
stack traceback:
	...om\mods\mc_worldmanager/realm/realmSchematicSaveLoad.lua:98: in function 'Load_Schematic'
	...om\mods\mc_worldmanager/realm/realmSchematicSaveLoad.lua:214: in function 'NewFromSchematic'
	...mes\Minetest_Classroom\mods\mc_worldmanager/commands.lua:308: in function 'func'
	...netest\minetest-5.5.1-win64\bin\..\builtin\game\chat.lua:79: in function <...netest\minetest-5.5.1-win64\bin\..\builtin\game\chat.lua:52>
	...st\minetest-5.5.1-win64\bin\..\builtin\game\register.lua:429: in function <...st\minetest-5.5.1-win64\bin\..\builtin\game\register.lua:415>
```
On rejoin, the realm appears in the realm list. When teleported to, however, it does not load.

**To Reproduce**
1. Join a MineTest world
2. Run the command `/realm schematic load MKRF512_all`
3. See server crash

**Expected behavior**
I expected a realm to be generated using the `MKRF512_all` schematic.

**Desktop:**
 - OS: Windows 10
 - MineTest Version 5.5.1

**Additional context**
- This also affects realm generation from the teacher tool
- Suspect this is an issue with just the `feature/schematicOffset` branch, since the line that causes the crash was added as part of #135. Also suspect that this is not schematic-specific, since the issue does not appear to be with the schematic files themselves

lukasgolson:
This PR adds a realm schematic config option to offset the placement of schematics. This has been used to fix the gap present in the MKRF maps.

Additionally, this PR fixes realms breaking when schematics are loaded into realms that are too small for the schematic; and exposes the realm clean method to command.


Lowie375 (dismissed):
Wall gaps in all of the `MKRF` schematics are fixed!

However, there are still some issues (primarily with the bottom `y`-layer of blocks) in some schematics:
- `MKRF_all`: bottom layer is air blocks
- `MKRF_slope`: bottom layer consists of unknown nodes - suspect the schematic may contain unregistered nodes beyond its borders that are being cropped out by the world borders, since realms made using this schematic use the black realm borders instead of the blue ones, and a blue border is present at the bottom of the realm area - we may want to mark the presence of these nodes as a different bug entirely, since it doesn't directly pertain to realm border placement
- `MKRF_aspect`: Same issue as `MKRF_slope`, except nodes clip outside of the realm borders. Suspect this is due to #139, in which case #140 should fix it. Regardless, unknown node presence in the schematic may want to be looked into
- `MKRF_dtm`: Same clipping issues as `MKRF_aspect`. However, the bottom layer of nodes is dirt (as expected)

lukasgolson:
> Wall gaps in all of the `MKRF` schematics are fixed!
> 
> However, there are still some issues (primarily with the bottom `y`-layer of blocks) in some schematics:
> 
> * `MKRF_all`: bottom layer is air blocks
> * `MKRF_slope`: bottom layer consists of unknown nodes - suspect the schematic may contain unregistered nodes beyond its borders that are being cropped out by the world borders, since realms made using this schematic use the black realm borders instead of the blue ones, and a blue border is present at the bottom of the realm area - we may want to mark the presence of these nodes as a different bug entirely, since it doesn't directly pertain to realm border placement
> * `MKRF_aspect`: Same issue as `MKRF_slope`, except nodes clip outside of the realm borders. Suspect this is due to #139, in which case #140 should fix it. Regardless, unknown node presence in the schematic may want to be looked into
> * `MKRF_dtm`: Same clipping issues as `MKRF_aspect`. However, the bottom layer of nodes is dirt (as expected)


Thanks for the review, I appreciate it. I think all these schematics can be cleaned by re-serializing the schematics using the realm system. I haven't had a chance to do it yet but I'll add it to the TODO list. 

* For `MKRF_all`, noted: I'll apply a -1 Y-offset.
* For `MKRF_slope`, this PR does not contain the cropping code at the bottom of the realm. I'll look into it a bit more as it might be a significant issue. Alternatively, it might be that the non-registered blocks are the old border present in the old schematics. The realm system generates its own border so, under this case, it shouldn't be an issue. Let me know if I'm mistaken about that.
* For `MKRF_aspect`, This PR does not pertain to this issue and it has been fixed in #139. Unknown nodes will be cleaned up when I re-serialize the schematics for the new format. I'll double check that it was the old border blocks that were giving errors.
* For `MKRF_dtm`, This PR does not pertain to this issue and it has been fixed in #139.

lukasgolson:
As an update, I've fixed the space at the bottom of all the realms by offsetting it by -1.

I've also determined that the realm walls are in-fact replacing the outer most block of the realms as @Lowie375 has described. I've determined that this is intended behavior to fit the realm within map chunks and so that these slightly larger (by 1 block) schematics match the full composite schematic `MKRF512_all`. 

If it's deemed that this is not how we'd like these particular schematics to behave, we just need to change each realms config file from: `schematic_size_x = 512` and` schematic_size_z = 512`  to `schematic_size_x = 513` and `schematic_size_x = 513`.

mods/mc_worldmanager/realm/realmSchematicSaveLoad.lua
```diff
@@ -94,11 +94,21 @@ end
 ---@return boolean whether the schematic fit entirely in the realm when loading.
 function Realm:Load_Schematic(schematic, config)
 
+
+    local schematicStartPos = self:LocalToWorldPosition(config.startOffset)
```
Lowie375: This line is currently causing a crash when generating realms from schematics via commands: see #144 for info

Lowie375 (dismissed):

mods/mc_worldmanager/realm/realmSchematicSaveLoad.lua
lukasgolson: Fixed

mods/mc_teacher/maps/MKRF512_aspect.conf
```diff
@@ -3,11 +3,14 @@ name = Malcolm Knapp Research Forest - aspect
...
 spawn_pos_y = 63
 spawn_pos_z = 161
-schematic_size_x = 513
+schematic_size_x = 512
 schematic_size_y = 175
-schematic_size_z = 513
-license = CC BY-SA 4.0
+schematic_size_z = 512
+offset_x = -1
+offset_y = -1
```
Lowie375: There is a 1-block high air gap on the bottom-most layer of the realm
```suggestion
offset_y = -2
```

mods/mc_teacher/maps/MKRF512_dtm.conf
```diff
@@ -3,11 +3,14 @@ name = Malcolm Knapp Research Forest - digital terrain model
...
 spawn_pos_y = 63
 spawn_pos_z = 161
-schematic_size_x = 513
+schematic_size_x = 512
 schematic_size_y = 175
-schematic_size_z = 513
-license = CC BY-SA 4.0
+schematic_size_z = 512
+offset_x = -1
+offset_y = -1
```
Lowie375: The bottom realm border seems to be clipping the lowest layer of blocks, which is most noticeable along the north edge of the realm
```suggestion
offset_y = 0
```

Lowie375 (dismissed):
Realm generation is no longer crashing, hooray!

The schematics look a lot better as well - there don't seem to be any more nodes clipping beyond the realm borders. That being said, I did notice a couple of things in some of them that should be resolved:

- `MKRF512_aspect`: There is now a 1-block high air gap on the bottom-most layer - suggest changing `offset_y` by `-1`
- `MKRF512_dtm`: The bottom realm border seems to be clipping the lowest layer of nodes in the realm - suggest changing `offset_y` by `+1`

Lowie375 (approved):
Looks good!

PR #138, opened 2022-07-23
Improve `mc_toolhandler` system

Lowie375:
* Reverse `mc_toolhandler` registration flow
  * Register tools from API function calls in other mods instead of from tool scrape in `mc_toolhandler`
  * Make `mc_toolhandler` a dependency instead of a dependent when registering tools
* Make `mc_toolhandler` put managed tools into toolbox by default (if `mc_toolmenu` is present)
* Add `inv_override` property to allow tools to be placed in different player inventory lists by default
* Replace various MineTest-defined helper functions in `mc_toolhandler` with `mc_helpers` equivalents
* Abstract out core of `get_player_item_location` function into `mc_helpers.getInventoryItemLocation`
* Make `list_info` local


mods/mc_toolhandler/mod.conf
```diff
@@ -1,4 +1,7 @@
 name = mc_toolhandler
 description = Provides functionality for giving/taking classroom tools from players
 depends = mc_helpers
-optional_depends = mc_teacher, mc_student, magnify, mc_tf, forestry_tools
+
+# add mods whose tools should be managed to the end of this comma-separated list
+# mc_toolmenu should ALWAYS be an optional dependency - do NOT remove it
+optional_depends = mc_toolmenu, mc_teacher, mc_student, magnify, mc_tf, forestry_tools
```
lukasgolson: At the end of my workday so hopefully I'm not misunderstanding what's going on here. Would the mc_toolhandler depend on the other mods as I think I'm reading it here or would it be the other way around?

E.g., mods that need have tools managed by mc_toolhandler should add it as a dependency rather than the other way around. At runtime, they should call the correct method to register their tools.

lukasgolson (commented):
I'm a bit confused about the mod.conf dependencies (which I have left a comment on). Otherwise, everything looks good to me!

mods/mc_toolhandler/mod.conf
Lowie375: > Would the mc_toolhandler depend on the other mods as I think I'm reading it here or would it be the other way around?

Yes, `mc_toolhandler` depends on the other mods.
The current registration system works by iterating through all of the registered tools at the time `mc_toolhandler` is initialized. It registers callbacks for any tools in the mods it depends on which contain either the `_mc_tool_privs` and `_mc_tool_include` fields in their definitions, and ignores everything else. Having the other mods as dependencies ensures that the mods whose tools are being registered have initialized all their tools before `mc_toolhandler` registers them.

All of the registration code is located in the `mc_toolhandler` mod rather than in the mods containing the tools, since that makes it easier to separate the tools from `mc_toolhandler` if needed. In the current implementation, the `_mc_tool_privs` and `_mc_tool_include` fields can just be ignored if `mc_toolhandler` is not installed or a mod is not added as a dependency. However, if it were the other way around (i.e. tools depend on `mc_toolhandler`), the tools would need to define a fallback for the registration function or risk a crash.

lukasgolson: Is it possible to do it the other way around? The current implementation sounds very error prone, a bit overcomplicated, and not very expandable -- especially by the way it forces us to manually modify the mod meta of an existing mod whenever creating a new mod. E.g., a third party would not be able to use your system easily.

Ideally, the other mods would call a single method in the public mc_toolhandler API to register itself with the system, and mc_toolhandler would take care of the rest. It wouldn't change where registration code is but would ensure that others who want to work with your mod don't need to make changes to your mod to make it work. 

While the way that I'm proposing would require for mods to check if mc_toolhandler is installed (see https://dev.minetest.net/Guide:_Optional_dependencies), it would be more intuitive and standard for registering tools with your system. Likewise, this would remove the issue of looping over all items registered with the game.  

Let me know your thoughts :)

Lowie375: I think I could refactor the system so that it works the other way around. Will look into starting that later today.

Lowie375: Have begun the refactor; will request a re-review once the refactor is complete

Lowie375:
Refactor is mostly complete! Will re-open the PR once support for tool groups has been re-implemented

Lowie375:
Have noticed a couple of issues that need to be fixed with `mc_toolhandler`; marking as draft until those have been fixed

lukasgolson (approved):
Looks great to me! The public facing API appears really easy to use which is a great plus!

pauldpickell (dismissed):
I am not sure which commit deleted it, but we should probably automatically check and ensure that all players join the game with the student notebook and the tutorial notebook and all teachers join with the student notebook, the teacher remote, and the tutorial notebook. It is not intuitive or self-evident how to get these tools if you have never played the game before yet they contain really important functionality to game play. We should make it stupid simple by putting it in their hands when they join. 

Also, I noticed that the tutorial recording tool is still automatically given from an on_joinplayer callback somewhere and is not registered in the toolhandler system yet. Is that forthcoming in the draft tutorial framework PR?

Lowie375:
> I am not sure which commit deleted it, but we should probably automatically check and ensure that all players join the game with the student notebook and the tutorial notebook and all teachers join with the student notebook, the teacher remote, and the tutorial notebook.

Noted! Will add a property that allows the default inventory for an item to be overridden. That should allow tools like the student notebook and teacher tools to always go to the "main" inventory

> Also, I noticed that the tutorial recording tool is still automatically given from an on_joinplayer callback somewhere and is not registered in the toolhandler system yet. Is that forthcoming in the draft tutorial framework PR?

Yep, that has been changed on the `features/tutorial_recording` branch. However, that branch is using an older version of `mc_toolhandler`, so I'll just register the tools on this branch and merge `develop` into `features/tutorial_recording` after this PR is pushed

pauldpickell (approved):
Works beautifully

PR #141, opened 2022-07-26
Feature/utm coordinates

### LINKED ISSUE #77: Feature: Real-world coordinates
In `map.conf` files for `Minetest_Classroom` realm schematics, it would be nice to store variables that describes the map projection parameters used to create real world maps. Since Minetest worlds use a cartesian coordinate system, it makes sense to support importing map schematics that are projected to UTM. Then, we can develop a simple position translator that converts Minetest coordinates to real UTM easting-northing and latitude-longitude pairs. To do this, we need to know the real-world coordinates (in the NAD83/WGS84 datum) for a location in the map schematic.

Solution: Create a position translator mod that uses the HUD to display the current real-world position in latitude/longitude and easting/northing units. Add to `map.conf` files the following variables:

- `UTM_zone` = `integer` in range of 1-60
- `UTM_map_origin`  = `table` with key-value pairs indicating the `easting` and `northing` values for the map schematic origin at `{x=0,z=0}`.

This small feature update would make it much easier to navigate and edit realms as well as implement new tools that could be dependent on this feature such as GPS. More practically, I see this feature being a critical asset for placing schematics and objects within realms and also finding and editing real-world locations.

lukasgolson:
This PR adds the feature request from #77.

- Adds config options to realm schematics to set the UTM zone and eastings/northings of the realm at origin -- local pos: (0,0).
- Adds a command to set the UTM zone and Eastings/Northings values of any existing realm.
- Adds a function to convert from UTM to lat/long pairs
- Adds a toggleable HUD to display player position in either world space, local space, grid space, UTM space, or Latitude/Longitude pairs.
- Adds a realm data command to set and retrieve the arbitrary realm storage values added onto realms 
- Added UTM info to the MKRF config files

### DIFF
--- a/mods/mc_helpers/init.lua
+++ b/mods/mc_helpers/init.lua
@@ -159,3 +159,7 @@ function mc_helpers.deepCopy(table)
     end
     return copy
 end
+
+function mc_helpers.round(x, n)
+    return tonumber(string.format("%." .. n .. "f", x))
+end
--- a/mods/mc_teacher/maps/MKRF512_all.conf
+++ b/mods/mc_teacher/maps/MKRF512_all.conf
@@ -7,3 +7,7 @@ schematic_size_x = 513
 schematic_size_y = 175
 schematic_size_z = 513
 license = CC BY-SA 4.0
+utm_zone = 10
+utm_north = true
+utm_origin_easting = 533244
+utm_origin_northing = 5464244
--- a/mods/mc_teacher/maps/MKRF512_aspect.conf
+++ b/mods/mc_teacher/maps/MKRF512_aspect.conf
@@ -7,3 +7,7 @@ schematic_size_x = 513
 schematic_size_y = 175
 schematic_size_z = 513
 license = CC BY-SA 4.0
+utm_zone = 10
+utm_north = true
+utm_origin_easting = 533244
+utm_origin_northing = 5464244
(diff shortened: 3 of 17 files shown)


lukasgolson:
> Here is an example set of UTM Zone 10N coordinates (in units of meters) to work with for all of the MKRF512 maps:
> 
> * Top (north of map): 5,464,756
> * Bottom (south of map): 5,464,244
> * Left (west of map): 533,244
> * Right (east of map): 533,756
> 
> Note that `Top - Bottom = 512` and `Right - Left = 512`, which are `(x,z)` the dimensions of the world (exclusive of the map barrier). If we include this information as optional input in the `map.conf` file, then we only need a single pair of tie coordinates (e.g., Top/Left or Bottom/Right) in order to define any other UTM coordinate in the map. The above coordinates are at the _center_ of those nodes at the defined `(x,z)` position, not the edge, so keep in mind when doing calculations, otherwise the calculations will be off by 0.5 m.

Hopefully this isn't a silly question, I have poor spatial understanding. I was thinking about this issue. Because all coordinates in the realm system, Minetest, and schematics are aligned at the center of the nodes, would the calculations still be off by 0.5m?

pauldpickell:
In that case, there should be no issue. Just wanted to flag that the local UTM coordinates should be calculated from node centers.

PR #146, opened 2022-07-27
Updated Tools

jocelynbachmann:
- Fixes missing clinometer icon
- Should fix disappearing HUD multiplayer bug for the compass and clinometer
- Separates and corrects outer bezel vs. shed rotation on the compass


mods/forestry-tools/clinometer.lua
```diff
@@ -101,86 +106,87 @@ local function update_hud(player, hud, hudname, text)
 end
 
 local function update_clinometer(player)
-	bg_angle = "0"
+	local pname = player:get_player_name()
+	instances[pname].bg_angle = "0"
```
pauldpickell, line 110: Given the discussion at our last meeting about Lua leaking memory with strings, I wonder if it would be best to cache this (and others) as an integer and then cast it to a string at the moment it is needed? Just thinking that this particular tool will be constantly cycling through strings as the player changes their viewing perspective.

mods/forestry-tools/clinometer.lua
```diff
@@ -101,86 +106,87 @@ local function update_hud(player, hud, hudname, text)
 end
 
 local function update_clinometer(player)
```
pauldpickell, line 108: Given the number of strings being modified with this function, I wonder if we should be be running `collectgarbage()` in here somewhere to release memory? Maybe @lukasdragon has some better ideas.

pauldpickell (approved):
Might consider how to deal with the string memory leak in Lua, specifically for the clinometer. I think we might open an issue separately for this so that we can have a standard way to mitigate any memory leaks.

PR #152, opened 2022-08-02
Dependency Patch for Startup Crash

jocelynbachmann:
Adds default and flowers as dependencies for mc_worldmanager to fix crash on startup (that is potentially only happening on mac)


lukasgolson (changes requested):
Tutorialbook.lua has merge markers that need to be resolve d before approval

PR #153, opened 2022-08-03
Fix/biomegen Frankenstein tree fix

### LINKED ISSUE #148: Biomegen decorator produces mismatched decorations
**Describe the bug**
Using the biomegen key for the decorator can create some weird results. Not sure if this is a bug of the biomegen mod or on our end, but I found some hybrid poplar-apple trees.

**To Reproduce**
1. `/realm new`
2. `/realm tp 2`
4. `/realm gen 2 v2 biomegen 10 123`
5. See result

**Expected behavior**
Should not see hybrid trees.

**Screenshots**
image

 - OS: Windows 10
 - Minetest: 5.5.1

lukasgolson:
This PR seems to fix #148 but further testing is required.

- Additionally, this PR enables height relative to sea-level as a determining factor of temperature in biome placement. E.g., higher elevations leads to a cooler temperature and snow on mountains. 
- This PR also adds a new solid black barrier block that is placed on the bottom of realms and can be built on. This has been added to improve aesthetics of realms that have large holes to the bottom of the realm, and to enable building in empty realms that would otherwise be unusable.


pauldpickell:
In testing, this does not fix #148 for me. Frankentrees are still there, so I am leaving the issue open, but merging this PR.

PR #157, opened 2022-08-04
Feature/biomegen "frankentree" fix

### LINKED ISSUE #148: Biomegen decorator produces mismatched decorations

lukasgolson:
This PR fixes #148 

I determined that the map generator was overlapping trees when there was little to no lag during generation. Out of pure speculation, I think this may have been because we were pulling random numbers too quickly in succession. The debug output was slowing down the code enough that this was not happening while trying to debug.

Because I was testing with larger maps than the default realm size for "efficiency" and/or my slow laptop, it was very rare for this to be replicated. As soon as I started working on my desktop, the bug became very apparent.

This PR fixes the issue by keeping track of where decorations are placed and ensuring that large decorations (e.g., trees) are not placed in the exact same location as each other.

### DIFF
--- a/mods/biomegen-v1.0/init.lua
+++ b/mods/biomegen-v1.0/init.lua
@@ -52,8 +52,6 @@ local function initialize(chulens, seaLevel, seed)
     nobj_heat = minetest.get_perlin_map(np_heat, chulens2d)
     nobj_heat_blend = minetest.get_perlin_map(noiseparams('mg_biome_np_heat_blend'), chulens2d)
 
-
-
     nobj_humid = minetest.get_perlin_map(np_humidity, chulens2d)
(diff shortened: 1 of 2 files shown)


pauldpickell (approved):
With the other changes to terrain generation, the `/realm gen 2 v2 biomegen 10 123` seed is no longer reproducible. But I cycled through a bunch of random seeds and found many different types of trees generated in the same world without being frankentrees, so I am satisfied that this fixes #148.

PR #158, opened 2022-08-04
Feature/elevation hud

### LINKED ISSUE #77: Feature: Real-world coordinates

lukasgolson:
This PR fulfils the following feature request https://github.com/ubc-minetest-classroom/Minetest_Classroom/issues/77#issuecomment-1199723446


mods/mc_worldmanager/hud.lua
```diff
@@ -8,21 +8,18 @@ local function createRealmInfoHudString(player)
...
-    })
-end
-
 function mc_worldManager.UpdateRealmHud(player)
     if (not mc_worldManager.hud:exists(player, "worldManager:currentRealm")) then
-        return false
+        mc_worldManager.hud:add(player, "worldManager:currentRealm", {
+            hud_elem_type = "text",
+            position = { x = 1, y = 0 },
+            offset = { x = -16, y = 5 },
```
pauldpickell: The position of these stacked HUD elements conflict with the minimap. Can we move everything to the bottom left instead? Or offset it enough that it appears below or beside the minimap?

pauldpickell (dismissed):
Looks and works great. We need to reposition the HUD elements for the current realm, coordinates, and elevation because it conflicts with the minimap when toggled on.

mods/mc_worldmanager/hud.lua
lukasgolson: I completely forgot that we had a minimap! This should be a pretty quick fix; I'll move it to the bottom left.

PR #160, opened 2022-08-08
Feature/biomegen biome selection

lukasgolson:
This PR adds support to biomegen biome decorator to specify the biome to use in generation. To complete these, the following changes have been introduced:
- Adds the `/realm biomes` command to print all registered biomes
- Adds passthrough support to parameters 6+ to the underlying generators for parameters to the `/realm gen` command
- Added support to the biomegen mod to use specified biomes during generation (and optional elevation chill).
- Numerous small fixes.

Example to use this PR: 

1. `/realm biomes` to display available biomes 
2. `/realm new` to create a new realm and view its ID
3a. `/realm gen <New Realm ID> v2 biomegen default random grassland`; where `grassland` is any biomekey.
3b. `/realm gen <New Realm ID> v2 biomegen default random 0.5 grassland`; where `0.5` is the elevation chill factor (vanilla default 0; forked biomegen default: 0.5); where `grassland` is any biomekey


pauldpickell:
Something weird is going on. It seems that regardless what biome key I submit, I seem to get a random biome. Following the `grassland` example in your PR, I get a forest:

image

Or a sandy beach with snow:

image

After looking at the code, I realized that your new `paramtable` is not behaving as expected, at least on my machine. In fact, I can type anything into the 5th parameter and it will generate something without throwing the expected ["not found" debug](https://github.com/ubc-minetest-classroom/Minetest_Classroom/blob/eaf6ca85bd6fd226f8c26437c117dc202b10f404/mods/biomegen-v1.0/init.lua#L74). For example, `/realm gen v2 biomegen foobar` returns a random biome every time.

mods/biomegen-v1.0/init.lua
```diff
@@ -63,6 +66,15 @@ local function initialize(chulens, seaLevel, seed)
 
     biomes = make_biomelist()
     decos = make_decolist()
+
+    if (biomes[_forcedBiome] ~= nil) then
+        forcedBiome = _forcedBiome
+    elseif (_forcedBiome ~= nil) then
+        forcedBiome = nil
+        Debug.log("[biomegen] Forced biome not found: " .. _forcedBiome)
```
pauldpickell, line 74: This is not working.

pauldpickell (dismissed):
Somehow, random biomes are being generated regardless of the input biome key for the command.

mods/biomegen-v1.0/init.lua
lukasgolson: That's very strange... When I execute `/realm gen 2 v2 biomegen default random desert` or `/realm gen 2 v2 biomegen default random 1 desert` the works as expected. Note that `default` is a parameter for `sea level` and random is a parameter for seed.

The biome is the 8th or 9th parameter for the command, respectively -- as per the `/realm help gen` command.

lukasgolson:
> Something weird is going on. It seems that regardless what biome key I submit, I seem to get a random biome. Following the `grassland` example in your PR, I get a forest:
> 
> image
> 
> Or a sandy beach with snow:
> 
> image
> 
> After looking at the code, I realized that your new `paramtable` is not behaving as expected, at least on my machine. In fact, I can type anything into the 5th parameter and it will generate something without throwing the expected ["not found" debug](https://github.com/ubc-minetest-classroom/Minetest_Classroom/blob/eaf6ca85bd6fd226f8c26437c117dc202b10f404/mods/biomegen-v1.0/init.lua#L74). For example, `/realm gen v2 biomegen foobar` returns a random biome every time.

`foobar ` in this case would be the parameter for seaLevel. Since `foobar` is not a number, seaLevel is using the default value of 30.

PR #162, opened 2022-08-08
Made pmeta and pdata in tutorial mod init.lua local

lukasgolson:
Minetest 5.6.0 crashes when pmeta and pdata in the tutorial mod init.lua file are global; this PR makes them local.


Lowie375:
Side note: this was patched on the `features/tutorial_recording` branch in commit [`65174f5`](https://github.com/ubc-minetest-classroom/Minetest_Classroom/commit/65174f5435df495c358f3e61d31598b414ffe652), so #129 should be largely unaffected by this patch

PR #165, opened 2022-08-10
Fix/realm gen set spawn crash

lukasgolson:
Weird Github duplicate branch shenanigans happening forcing a duplicate PR to receive the most up-to-date commit history.

This PR fixes a rare crash that happens when using the DNR heightmap generator when the old spawn point was outside of the terrain generation bounds. 
- Fixed crash caused by trying to use non-existent heightMapTable entries when setting realm spawn during terrain generation.
- Fixed potential incorrect heightMap values if realmFloorLevel is somehow above terrain.
- Updated setSpawn command to ensure that it always sets valid in-realm spawnpoints.
- Added a 2-step spawnpoint fallback when generating new terrain (first fallback step is to try the terrain height at the center of the realm, if this fails, we set the spawnpoint to the center of the realm, 5 blocks below the upper boundary.


pauldpickell (approved):
Looks good

PR #167, opened 2022-08-12
Updated student notebook functionality

jocelynbachmann:
"Join Classroom" button should now show a list of all realms a student created or has been to, with "Join" teleporting to the selected realm. "Register" currently does nothing as access codes are not implemented yet.

"My Coordinates" button should list all saved coordinates in the format "RealmName UTM, LatLong, Note". If unavailable for the realm, UTM coordinates are replaced by world (xyz) coordinates, and latlong coordinates are excluded. Coordinate entries can now also be deleted individually.


mods/mc_student/init.lua
```diff
@@ -265,38 +348,43 @@ minetest.register_on_player_receive_fields(function(player, formname, fields)
...
+				selectedCoord = event.index
+			end
 		elseif fields.go then
-			local pname = player:get_player_name()
-			local pmeta = player:get_meta()
-			if not context.selected then
-				context.selected = 1
+			if selectedCoord ~= 0 then
+				local new_pos = minetest.deserialize(pmeta:get_string("coordinates")).coords[selectedCoord]
+				player:set_pos(new_pos)
```
pauldpickell: There should be a check here first to ensure that the realm still exists (has not been deleted) and that the student is still authorized to join the realm (i.e., either registered in the realm or it is a spawn realm or it is a default realm). Want to avoid allowing students to get into realms they shouldn't be in and also teleporting them to an empty location. See comment above at line 179.

mods/mc_student/init.lua
```diff
@@ -69,132 +115,157 @@ local function show_report(player)
...
-			mc_student_coordinates[#mc_student_coordinates + 1] = pxyz[i] .. " " .. pnotes[i] .. ","
+		local prealms = pdata.realms
+		local pcoords = pdata.coords
+		local pnotes = pdata.notes
+
+		if pcoords then
+			for i in pairs(pcoords) do
+				local realm = Realm.GetRealm(prealms[i])
+				local pos = pcoords[i]
+				local utm = realm:WorldToUTMSpace(pos)
```
pauldpickell: When the realm is not defined (e.g., it was deleted) then the game crashes when the notebook is accessed.

`AsyncErr: Lua: Runtime error from mod 'mc_student' in callback on_playerReceiveFields(): ...assroom-update-student-notebook\mods\mc_student\init.lua:179: attempt to index local 'realm' (a nil value)`

You can reproduce this yourself with the following:

- Start new game
- `/realm new`
- `/realm tp 2`
- Record a coordinate anywhere in the realm
- `/realm delete 2`
- Access the student notebook

mods/mc_student/init.lua
pauldpickell: Also, using noclip, I was able to move outside the physical bounds of the realm and still record a coordinate and then teleport to that coordinate. Can we use map.conf data to limit the logical coordinates? Just trying to avoid this tool being used nefariously since it allows you to teleport anywhere in the world.

mods/mc_student/init.lua
```diff
@@ -69,132 +115,157 @@ local function show_report(player)
...
+				local latlong = realm:WorldToLatLongSpace(pos) 
+				local entry = realm.Name .. " "
+
+				if utm then
+					entry = entry .. math.floor(utm.x) .. "E " .. math.floor(utm.z) .. "N " .. math.floor(utm.y) .. "Z"
```
pauldpickell: Suggest changing the "Z" to "Elev" or "m" to make it more apparent this value is referring to elevation.

mods/mc_student/init.lua
```diff
@@ -69,132 +115,157 @@ local function show_report(player)
...
+				else
+					entry = entry .. "x=" .. math.floor(pos.x) .. " y=" .. math.floor(pos.y) .. " z=" .. math.floor(pos.z)
+				end
+
+				if latlong then
```
pauldpickell: Can we augment this so that if there are no UTM or latlong coordinates defined for the realm in map.conf, then it defaults to a format of `x=x , y=y, z=z`?

pauldpickell (changes requested):
This is a nice improvement. There needs to be further validation of coordinates to ensure that this tool is not abused by a malicious client.

mods/mc_student/init.lua
jocelynbachmann: Does this not currently do that (I haven't been able to test cases where the realm doesn't have utm/latlong due to mac issues)? Or are you just referring to adding the commas between the x y and z?

mods/mc_student/init.lua
```diff
@@ -265,38 +348,43 @@ minetest.register_on_player_receive_fields(function(player, formname, fields)
...
-		    	local new_pos_char = temp.coords[context.selected]
-			local new_pos_tab = pos_split(new_pos_char)
-		    	player:set_pos(new_pos_tab)
-		elseif fields.deleteall then
-			local pmeta = player:get_meta()
+		elseif fields.delete then
+			local data = minetest.deserialize(pmeta:get_string("coordinates"))
+			local newCoords, newNotes, newRealms = {}, {}, {}	
+
+			for i,coord in ipairs(data.coords) do
```
pauldpickell: This can crash the game when the coordlist is empty and the player clicks the `Delete` button.
Also, the when there is only one item (coordinate) in the list, the `Delete` button will not remove it and only the `Clear All` button removes the coordinate.

`AsyncErr: Lua: Runtime error from mod 'mc_student' in callback on_playerReceiveFields(): ...assroom-update-student-notebook\mods\mc_student\init.lua:376: attempt to index local 'data' (a nil value)`

mods/mc_student/init.lua
pauldpickell: No, currently the format is `xE zN yZ, y.yyyy°N x.xxxx°W`.

mods/mc_student/init.lua
```diff
@@ -183,7 +183,7 @@ local function show_coordinates(player)
 					local entry = realm.Name .. " "
 
 					if utm then
-						entry = entry .. math.floor(utm.x) .. "E " .. math.floor(utm.z) .. "N " .. math.floor(utm.y) .. "Z"
+						entry = entry .. math.floor(utm.x) .. "E " .. math.floor(utm.z) .. "N Elev: " .. math.floor(utm.y)
```
pauldpickell: Suggest `entry = entry .. math.floor(utm.x) .. "E " .. math.floor(utm.z) .. "N Elev: " .. math.floor(utm.y) .. "m"`

pauldpickell (approved):

PR #169, opened 2022-08-19
Feature/trees

lukasgolson:
This PR marks the introduction of the dynamic tree proof-of-concept framework. Note that while this is still in very early stages of development, it's still lots of fun to play around with.


pauldpickell:
Admittedly, this is a lot of fun and I can see a lot of potential for this. Looking forward to how this develops. I guess we will need to develop many new L-system trees or crowd-source them.

PR #185, opened 2022-12-22
Add notebook formspec + helper docs to `mc_core`

Lowie375:
* Add new `gui.lua` file to `mc_core` for formspec/GUI templates
  * Currently contains [notebook formspec template](https://github.com/ubc-minetest-classroom/minetest_classroom/blob/features/tutorial_recording/mods/mc_tutorial/init.lua#L133) from [features/tutorial_recording](https://github.com/ubc-minetest-classroom/minetest_classroom/tree/features/tutorial_recording)
* Add README with helper function documentation to `mc_core`

### DIFF
--- a/mods/mc_core/README.md
+++ b/mods/mc_core/README.md
@@ -1,0 +1,292 @@
+# mc_core
+
+Core utilities for Minetest Classroom
+
+## Global Objects
+
+### `minetest_classroom` (*`table`*)
+
+Global table used by various mods to store Minetest Classroom data
+
+## Bundled Utilities
+
+- `bit.numberlua` ([`numberlua.lua`](numberlua.lua)): bitwise operators in Lua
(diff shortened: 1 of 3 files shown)


pauldpickell:
This will make it easy to add additional content for the student notebook and teacher controller. Makes me think: do we really need a separate tool to access the tutorials or could we just as easily add it as another tab on the student notebook?

Lowie375:
> Makes me think: do we really need a separate tool to access the tutorials or could we just as easily add it as another tab on the student notebook?

That's a good point, since it would make sense to have that functionality in the student notebook. It'll be worth looking into doing that after the general tutorial recording system improvements have been merged, since it may take some effort to shift things around and get everything working properly.

If that were to happen, it might be worth adding another access point for tutorial editing, since teachers might not think to look in the student notebook to do that.

PR #222, opened 2023-06-14
Marker functionality patch

### LINKED ISSUE #220: HUD marker does not display correctly for players in a classroom
**Describe the bug**
When a player records a coordinate in a classroom and then clicks the place a marker button, the HUD marker appears only for the player who placed the marker.

**To Reproduce**
1. Open student notebook.
2. Navigate to Map tab.
3. Type a message and click Save Current Coordinates.
4. Highlight the new message in the list and then click the place marker in world button.
5. Observe that other players cannot see the HUD marker that is being shared.

**Expected behavior**
All other players in the classroom should be able to see the HUD marker. Need to take care that `minetest.register_on_joinplayer` and `minetest.register_on_leaveplayer` are set up to manage these HUD markers for players that are coming and going as well as when players teleport between classrooms. 

 - Version 5.7.0

Lowie375:
Makes markers work as originally intended, (hopefully) fixing #220

### DIFF
--- a/mods/mc_core/coordinates.lua
+++ b/mods/mc_core/coordinates.lua
@@ -1,74 +1,106 @@
 function mc_core.record_coordinates(player, message)
-	if mc_core.checkPrivs(player, {interact = true}) then
-		local pmeta = player:get_meta()
-		local pos = player:get_pos()
-		local realmID = pmeta:get_int("realm")
-		local temp, clean_message
-		clean_message = minetest.formspec_escape(message)
-		temp = minetest.deserialize(pmeta:get_string("coordinates")) or {realms = {}, coords = {}, note_map = {}}
+    if mc_core.checkPrivs(player, {interact = true}) then
+        local pmeta = player:get_meta()
+        local pos = player:get_pos()
+        local realmID = pmeta:get_int("realm")
+        local temp, clean_message
+        clean_message = minetest.formspec_escape(message)
+        temp = minetest.deserialize(pmeta:get_string("coordinates")) or {realms = {}, coords = {}, note_map = {}}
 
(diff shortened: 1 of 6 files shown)


mods/mc_core/coordinates.lua
```diff
@@ -1,74 +1,106 @@
 function mc_core.record_coordinates(player, message)
-	if mc_core.checkPrivs(player, {interact = true}) then
-		local pmeta = player:get_meta()
-		local pos = player:get_pos()
-		local realmID = pmeta:get_int("realm")
-		local temp, clean_message
-		clean_message = minetest.formspec_escape(message)
-		temp = minetest.deserialize(pmeta:get_string("coordinates")) or {realms = {}, coords = {}, note_map = {}}
+    if mc_core.checkPrivs(player, {interact = true}) then
```
pauldpickell, line 2: This should be checking for `shout = true`, too.

pauldpickell (changes requested):

mods/mc_core/coordinates.lua
Lowie375: Right, thanks for catching that!

Lowie375: I've added the priv check to the callbacks for the "mark" and "share" buttons rather than the coordinate recording function (this one).
People without `shout` should still be able to record coordinates if they'd like, but they should not be able to share the notes associated with them.

Lowie375:
Apologies for all the whitespace changes introduced - they're mainly to clean up file formatting. There's a viewing toggle for them in the "Files changed" tab, under the settings icon.

pauldpickell (approved):

Issue #229, opened 2023-07-18
Feature Request: Delay Display of IP Whitelist


pauldpickell:
When the server tab is clicked in the teacher controller, there is a significant delay as it takes time to return all the IP addresses. This also causes lag/delay whenever other actions are taken on the server tab, like exiting from the ban manager. It might be helpful to only display the full IP list when requested. Maybe we can add a button to show it in a popup?

Lowie375:
Whitelist has been moved from the main server tab to a popup on `features/unified_gui_ii` as of [`4ff019f`](https://github.com/ubc-minetest-classroom/minetest_classroom/commit/4ff019f16b1990a78522a728eb9b158d7dd4c9a2), which should reduce lag when opening the server tab. However, that doesn't fix the fundamental issue.
Will add a button to show the whitelisted IPs to the popup - we may also want to consider some other ways to condense the displayed IPs for readability and better load times

pauldpickell:
Solved by #231

Issue #228, opened 2023-07-18
Feature Request: Handle Realm Deletion Asynchronously


pauldpickell:
Currently realms are deleted immediately when triggered from the teacher controller. Players are teleported out of the realm to the spawn realm and then the realm is immediately removed and the coordinate space is freed up. This is all expected behavior, however it causes significant lag on the server.

Realm deletion should be queued and then either done asynchronously or triggered by a teacher when it is not expected to impact the game session. A few things need to be implemented for this to work properly:

- [x] Delay realm deletion by at least 30 seconds after teleporting the player. This ensures that the players have been removed and their clients have time to get the nodes and chunks of the spawn realm.
  - *Nanjou: latest version has a 15 second delay, but changing this is trivial if 15 seconds isn't enough time*
- [x] Add a property/attribute to realms that indicates if the realm is hidden or not (i.e., queued for deletion). 
- [x] Add three tabs where the current "Available Classrooms" list is under the classrooms tab of the teacher controller: "Public", "Private", "Hidden". 
- [x] Add two tabs where the current "Available Classrooms" list is under the classrooms tab of the student notebook: "Public" and"Private".
- [x] Under the "Public" tab: list the classrooms of type spawn and classroom; teachers see buttons to Teleport, Edit, and Hide in the teacher controller; students see button to Teleport in the student notebook.. 
- [ ] Under the "Private" tab: students see a list of all their own instanced classrooms in the student notebook and buttons to "Teleport" and "Hide"; somewhere else in the student notebook formspec are some simple features to create an instanced realm (students are limited to one instanced realm by default - maybe this is another setting that the teacher can change in the teacher notebook?); teachers see a list of their own instanced classrooms in the student notebook with buttons to "Teleport" and "Hide"; teachers see all instanced classrooms created by all players in the teacher controller with buttons to "Teleport" and "Hide".
- [x] Under the "Hidden" tab (teacher controller only): teachers see all realms queued for deletion and buttons for "Teleport", "Delete Selected Now", "Delete All Now", and "Restore". Restoring the realm, moves it from the list in the "Hidden" tab to the "Public" or "Private" tabs, depending on the type of realm it is. When either of the delete buttons are activated, a popup should be displayed warning the teacher that deleting classrooms is irreversible, may take a long time to complete, and the server may become unresponsive in the meantime. Maybe the delete buttons are only made available to a player who has both teacher and admin privs?

pauldpickell:
solved by #231

PR #223, opened 2023-06-16
Unified GUI updates, part 2

Lowie375:
Continuation of #209: current goal is to complete the teacher tool's redesign
* Design + implement remaining tabs of teacher controller (Help, Reports, Server)
* Reformat report log + store in `mc_teacher`
* Abstract out various helper functions
* Make indent style (tabs/spaces) consistent in various files

### ROADMAP

- Overview
  - [ ] Make images for Moderation, Reports, and Server tabs
  - [x] Ensure "edit server rules" button is functional
- Classrooms
  - [x] Implement "edit" button
  - [x] Add option for terrain generation + decoration
  - [ ] Populate BG music and skybox options
  - [ ]  #171
- Map
  - [x] Implement "teleport all" button
  - [ ] Add "save" button texture
  - [ ] #184 
- Players
  - [ ] Implement grouping features
  - [ ] Implement actions
    - [x] teleport/bring
    - [x] freeze
    - [ ] audience
  - [x] Implement role buttons (none/student/teacher/admin)
- Moderation
  - [ ] Add "send" button texture
  - [x] Make "mute" button functional
  - [ ] Add anti-spam filter for DMs and chat messages
- Reports
  - [x] Design formspec
  - [x] Create system for viewing logged reports
  - [x] Notify teachers about reports sent while offline
  - [x] Allow reports to be deleted
- Help
  - [x] Design formspec
  - [ ] Add secondary panel with Minetest Classroom setup help
- Server
  - [x] Redesign formspec
  - [x] Implement formspec:
    - [x] Server shutdown + scheduling
    - [x] Global messenger
    - [x] IP whitelist
    - [x] Ban manager

PR #231, opened 2023-08-01
Unified GUI updates, part 3

### LINKED ISSUE #227: Instanced realms cause incorrect context on student formspec list
**Describe the bug**
When an instanced realm is created by a teacher player, the context for other players is incorrect. When selecting classrooms to teleport to from the student notebook, the selected realm and teleported realm are not the same.

**To Reproduce**
1. Create an instanced realm as teacher player
2. Another player joins the server
3. The other player opens the student notebook and navigates to the classrooms tab
4. The other player selects a classroom from the list and clicks the teleport button, but is brought to another classroom

**Expected behavior**
Instanced classrooms of a player should not influence the context of other players when selecting classrooms.

Lowie375:
This PR includes a bunch of fixes for identified issues with the current GUI, as well as various new GUI-related features and feature implementations.

Notable inclusions:
* Fix #227
* Implement #229
* Make formspec `textlists` use absolute references to players (per addendum to #223)
* Implement Freeze (player) and Edit (classroom) buttons
* Change classroom viewer to tabbed design, separating Public, Private, and Hidden classrooms
  * Any classroom can be hidden, but only hidden classrooms can be deleted
  * Implement #228 (mostly)
* Change "Instanced" realm type to "Private"
* Add more popups for potentially dangerous actions
  * Kick, ban, server restart
* Properly handle privilege granting/revoking via server role assignment buttons and realm privilege application
  * Add `mc_core` helpers to call `on_priv_grant` and `on_priv_revoke` callback functions
* Track length of `networking` IP whitelist in mod metadata
* Add new sticky note GUI template for popups
* Make various tweaks to overall GUI formatting


Lowie375:
Some features are still incomplete: these will be added in a later PR (which should hopefully come soon!)
- [ ] Instanced realm creation for students
- [ ] Player groups
- [ ] Remaining action buttons
  - [ ] audience
  - [ ] timeout
- [ ] Auto-granting the `fly` privilege when the `noclip` privilege is granted
- [ ] Classroom generation improvements
  - [ ] Custom biome definitions
  - [ ] Manual UTM/lat-lon coordinate entry for non-twin realms
  - [ ] Support for LiDAR twinning
- [ ] Redesign of `mc_tutorial` popups

pauldpickell:
The new tab system for classrooms public/private on the student notebook creates an interesting wormhole where I can move a classroom to hidden as teacher, teleport there, and then return to student notebook to find I belong to no classroom. This is intended behaviour and does not seem to cause any issue, though it might have unintended consequences/confusion.

pauldpickell:
When I edit the spawn classroom and change it to private, then set another classroom as spawn, I still cannot hide the original spawn classroom. But I can hide any other classroom that I designate as spawn.

- [x] Disallow more than one classroom set as spawn; overwrite when necessary in edit mode and change the previous spawn classroom type to 'classroom' as a fall back
- [ ] Disallow changing the spawn classroom to private; I was not able to test this in singleplayer mode, but this could prevent non-teacher players from joining/spawning correctly

pauldpickell (changes requested):
Changes needed to deal with unexpected state transitions for spawn realm type.

Lowie375:
> When I edit the spawn classroom and change it to private, then set another classroom as spawn, I still cannot hide the original spawn classroom. But I can hide any other classroom that I designate as spawn.

I suspect this is because setting a classroom's type to "spawn" doesn't actually change which classroom is recognized as the spawn in `mc_worldmanager` (which manages classroom teleportation). Will work on making the requested changes over the next few days

Lowie375:
> The new tab system for classrooms public/private on the student notebook creates an interesting wormhole where I can move a classroom to hidden as teacher, teleport there, and then return to student notebook to find I belong to no classroom.

A couple options worth considering:
- Add a "hidden" tab to the student notebook for students currently in a hidden realm (with only their current realm shown), then hide that tab once those student leave the realm
- When a realm is hidden, send a message to everyone currently in the realm: everyone gets told that the realm has been hidden, students also get told that they won't be able to re-enter the realm once they have left it

pauldpickell:
> > The new tab system for classrooms public/private on the student notebook creates an interesting wormhole where I can move a classroom to hidden as teacher, teleport there, and then return to student notebook to find I belong to no classroom.
> 
> A couple options worth considering:
> 
>     * Add a "hidden" tab to the student notebook for students currently in a hidden realm (with only their current realm shown), then hide that tab once those student leave the realm
> 
>     * When a realm is hidden, send a message to everyone currently in the realm: everyone gets told that the realm has been hidden, students also get told that they won't be able to re-enter the realm once they have left it

I think another option is to just check that the realm is empty of all players, then allow it to be hidden. That would solve any issue of a player being in a hidden realm. Teachers can always move players out of realms if needed. Alternatively, we could just force teleport players out to the spawn realm when a realm is marked as hidden, in the same way that happens when realm deletion is initiated.

pauldpickell:
I just teleported to a realm on the hidden tab, which moved my player to that realm, but when I closed the teacher controller GUI, it instantly teleported me back to spawn. Now when I open the classroom tab on the teacher controller, I can flip between spawn and hidden realms and it shows `(1 player)` for both. Then, I tried teleporting back to the realm again, and it returned me to spawn every time. So I tried another realm that I had hidden, and it seems I was able to successfully move around without being teleported back to spawn. However, the hidden tab formspec list shows `(1 player)` now for both realms. 

Then, I decided to create a new empty realm and moved it to hidden for testing. Once I did that, the realm I was trying to teleport to completely disappeared from the list. After some thought, I realized that this realm I was trying to teleport to had probably been deleted when I was testing the PR, but it was not removed from the list (difficult to tell because all realm names are the same). So realms deleted may not be removed from the hidden list.

Lowie375:
> * [ ] Disallow changing the spawn classroom to private; I was not able to test this in singleplayer mode, but this could prevent non-teacher players from joining/spawning correctly

Instead of doing this, I've added some additional checks on join that make sure players have access the the realm they were in when they left the game.
If they can't rejoin their previous realm (due to realm category changes, for instance), they will be moved to the spawn realm (which is joinable by all players, and will always exist, since a new one is generated if the existing spawn realm's category is changed)

Lowie375:
> I think another option is to just check that the realm is empty of all players, then allow it to be hidden. That would solve any issue of a player being in a hidden realm. Teachers can always move players out of realms if needed. Alternatively, we could just force teleport players out to the spawn realm when a realm is marked as hidden, in the same way that happens when realm deletion is initiated.

I'm going to do something a bit between these two approaches: if someone tries to hide a realm with players inside of it, they will be shown a popup asking if they want to hide the realm and kick out any players currently in it or cancel the hide action.

pauldpickell:
Some additional testing with hosted server reveals that the textlist of connected players in the players tab of the teacher controller is not reloaded when you open/close the tool. The only reload I have observed is when you select different tabs (students/teachers/classrooms). The effect is that stale player names remain in the textlist context.

To reproduce:
- Start a hosted server game
- Open another client and connect to 127.0.0.1 as `student1`
- Open the teacher controller and observe `student1` in the textlist, close the controller
- Leave the game from the other client and reconnect as `student2`
- Open the teacher controller and observe `student1` in the textlist
- Change the tab to teachers or classroom and then back to students
- Observe `student2` in the textlist

pauldpickell:
Another unexpected behavior: join a hosted server as a student and you are given tutorial notebook and student notebook tools. Leave the game and then rejoin and both tools are taken from inventory without applying any priv changes from teacher controller.

Edit: appears that `student` priv is somehow revoked when rejoining the server.
Edit2: granting the player the `student` priv from the teacher controller works as expected.

pauldpickell:
> Instead of doing this, I've added some additional checks on join that make sure players have access the the realm they were in when they left the game. If they can't rejoin their previous realm (due to realm category changes, for instance), they will be moved to the spawn realm (which is joinable by all players, and will always exist, since a new one is generated if the existing spawn realm's category is changed)

I just tested this by creating a realm, teleporting another player there, leaving the game with that player, then hiding the realm, and rejoining, and the player is able to rejoin the realm that they left the game in, which is now hidden. Expected behavior is that the player's realm is checked if it can be joined and if not then teleported to the spawn realm. 

I also tested by changing the realm type to private, but this resulted in damage to the player on rejoining, dying, and then spawning into the spawn realm. Again, the player should just automatically be teleported to spawn instead of dying.

pauldpickell:
Freeze/unfreeze works great except when the player is frozen in a classroom that is hidden. When this happens, the player is gets the usual chat message that the classroom has been hidden and you were brought back to the spawn. But the player is not automatically teleported to the spawn realm, instead dies, and then when they do respawn in the spawn realm, they are not frozen. This causes the unexpected behavior that the player can seem to move around, but cannot teleport to any classrooms and from the teacher controller the player still shows as having the frozen state. 

Also, we might think about adding a chat message or something that indicates to the player that they are frozen.

pauldpickell:
All terrain generators behave as expected except DNR. DNR was meant to be map regen method and so it relies on some nodes already being defined in the realm. Apparently, if the realm is generated with DNR, the behavior is that the entire realm is filled with dirt and then the player spawns outside of the bounds of the realm, which results in a loop of damage and respawning outside the realm. For this reason, DNR should be removed from the terrain generator list. 

Regen methods should be moved to the edit classroom popup. In other words, preserve the current realm settings and simply change the nodes using the terrain generators and decorators. The seed, sea level, generator, and decorator are all stored in `realm.MetaStorage`.

pauldpickell (changes requested):
Five issues raised with these changes:

- [x] 1. Context of the player textlist is stale when player joins/leaves/rejoins
- [x] 2. Players join hosted server receive `student` priv, then lose it when they leave/rejoin
- [x] 3. Improper handling of cases where a player leaves, has their realm hidden or made private, and they rejoin with unexpected behavior
- [x] 4. Freeze/unfreeze fails when a player is frozen in a realm that is hidden.
- [x] 5. DNR and other regen methods need to be moved to classroom edit popup

See comments for more details.

Lowie375:
> * [ ] 2. Players join hosted server receive `student` priv, then lose it when they leave/rejoin

Suspect this may be happening because `student` isn't being granted universally, that should be a simple fix

> * [ ]  3. Improper handling of cases where a player leaves, has their realm hidden or made private, and they rejoin with unexpected behavior
> * [ ]  4. Freeze/unfreeze fails when a player is frozen in a realm that is hidden.

This should be able to be fixed with some more checks, since the current checks are only using `realm:getCategory().joinable()` - will change behaviour for hidden realms and frozen players. Will also make sure players remain frozen on respawn (and that any other similar effects get applied on respawn)

> [ ] 5. DNR and other regen methods need to be moved to classroom edit popup

Will remove DNR for now. It'll get added back into the edit popup once the infrastructure for regenerating realm terrain via the GUI is in place

mods/mc_core/freeze.lua
```diff
@@ -0,0 +1,91 @@
...
+	return pmeta and minetest.deserialize(pmeta:get("mc_core:frozen")) or false
+end
+
+---@public
+---Runs a function on player, temporarily unfreezing them while the function runs if they are frozen
+---Arguments after the function definition will be passed into in, similar to minetest.after 
+---The player argument will not be passed into func by default, so it should be added to the argument list in the appropriate position
+---@param player ObjectRef to treat as unfrozen
+---@param func Function to run
+function mc_core.run_unfrozen(player, func, ...)
```
pauldpickell, line 88: Can this not be a more general `mc_core.pass_function` function? Just thinking that the naming and use is not intuitive. For example, can we not just have another `mc_teacher.run_unfreeze` function that is passed to this `mc_core.pass_function`? This would make the use of this "core" function much more flexible.

pauldpickell (approved):

mods/mc_core/freeze.lua
Lowie375: Theoretically, this could be made more general in the future, but I think that may be a bit too much of an abstraction.
The main reason that this is its own function is because it's using the local `freeze_player` and `unfreeze_player` functions, which don't alter a player's `mc_core:frozen` metadata attribute (which is used in `mc_core.is_frozen`).
If someone wanted to do something similar with other functions, they could use a check before the inner function (`func`), and call the appropriate outer functions before and after the inner function if that check passes. Abstracting that out into a theoretical `mc_core.pass_function` function would likely create a clunky interface, since `mc_core.pass_function` would have to handle all of the involved functions and their arguments (which can vary between functions)

Lowie375: > Just thinking that the naming and use is not intuitive.

Noted, will try to improve that in the future

